PR #20, opened 2018-02-22
Add import support for random id and random integer

danawillow:
Allow importing random resources to be able to replace a config value with a value interpolated from the random provider without experiencing diffs.

See https://github.com/terraform-providers/terraform-provider-google/issues/1054 (which then unblocks https://github.com/terraform-providers/terraform-provider-google/issues/780) for more information on why this is a useful feature.

Once this is merged, I'll add import support for the remaining random resources for completeness.

### DIFF
--- a/random/resource_integer_test.go
+++ b/random/resource_integer_test.go
@@ -20,6 +20,12 @@ func TestAccResourceIntegerBasic(t *testing.T) {
 					testAccResourceIntegerBasic("random_integer.integer_1"),
 				),
 			},
+			{
+				ResourceName:      "random_integer.integer_1",
+				ImportState:       true,
+				ImportStateId:     "3,1,3,12345",
+				ImportStateVerify: true,
+			},
 		},
 	})
 }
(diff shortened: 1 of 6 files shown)


random/resource_integer_test.go
```diff
@@ -20,6 +20,12 @@ func TestAccResourceIntegerBasic(t *testing.T) {
 					testAccResourceIntegerBasic("random_integer.integer_1"),
 				),
 			},
+			{
+				ResourceName:      "random_integer.integer_1",
+				ImportState:       true,
+				ImportStateId:     "3,1,3,12345",
+				ImportStateVerify: true,
+			},
```
paddycarver: Ideally, we'd see a version of this that didn't set the Seed, just to exercise that code path, too.

danawillow: That's a difficult test to write: the import tests create a resource and then import it and check that the result matches. How would you go about guaranteeing that the resource will be created in that first step with a particular result without the seed?

paddycarver: That's a fair question! Request rescinded.

PR #22, opened 2018-04-03
Add minimum character constraints to random_string

phekmat:
The motivation behind this is to more reliably create passwords that will pass complexity requirements

### DIFF
--- a/random/resource_string.go
+++ b/random/resource_string.go
@@ -2,7 +2,10 @@ package random
 
 import (
 	"crypto/rand"
+	"math/big"
+	"sort"
 
+	"github.com/hashicorp/errwrap"
 	"github.com/hashicorp/terraform/helper/schema"
 )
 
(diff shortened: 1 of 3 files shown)


mildwonkey:
Thanks for adding this feature :)

mildwonkey:
@kitsunde I'm sorry to hear about that and thanks for sharing the workaround. I'm certain it wasn't the intent of this PR - I will take a look at it.

phekmat:
Apologies all. @mildwonkey I'm still relatively new with terraform providers, but I think the correct approach is to use a state migration to add the defaults to the instance state. I've opened a PR with that change (#29)

mildwonkey:
Thank you @phekmat - I opened #28 just a few minutes before you with basically the same function. (I'm also new to this!) I'll close my PR; it was a good learning experience for me but we should let you own the fix.

ChrisAnn:
This hit us too and we had a short production outage as a result :sob:

Issue #26, opened 2018-05-22
random selections not produced for small sets or fast operations


wadeholler:
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version
Terraform v0.11.1
+ provider.random v1.3.0

### Affected Resource(s)
resource_shuffle or all random resources

### Terraform Configuration Files
```hcl
resource "random_shuffle" "mu_random_subnet" {
  input = ["a", "b", "c"]
  result_count = 1
  count = "10"
}

output "mu_shuffles" {
  value = ["${random_shuffle.mu_random_subnet.*.result}"]
}

```

### Debug Output
```bash
...
random_shuffle.mu_random_subnet[2]: Creation complete after 0s (ID: -)
random_shuffle.mu_random_subnet[0]: Creation complete after 0s (ID: -)
random_shuffle.mu_random_subnet[6]: Creation complete after 0s (ID: -)
random_shuffle.mu_random_subnet[9]: Creation complete after 0s (ID: -)
random_shuffle.mu_random_subnet[4]: Creation complete after 0s (ID: -)
random_shuffle.mu_random_subnet[1]: Creation complete after 0s (ID: -)
random_shuffle.mu_random_subnet[7]: Creation complete after 0s (ID: -)
random_shuffle.mu_random_subnet[3]: Creation complete after 0s (ID: -)
random_shuffle.mu_random_subnet[5]: Creation complete after 0s (ID: -)
random_shuffle.mu_random_subnet[8]: Creation complete after 0s (ID: -)

Apply complete! Resources: 10 added, 0 changed, 0 destroyed.

Outputs:

mu_shuffles = [
    [
        a
    ],
    [
        a
    ],
    [
        a
    ],
    [
        a
    ],
    [
        a
    ],
    [
        a
    ],
    [
        a
    ],
    [
        a
    ],
    [
        a
    ],
    [
        a
    ]
]
```

### Expected Behavior
Random results should have been produced

### Actual Behavior
This is not random for small or fast operations

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply`

### Suggestion

I have tested using UnixNano() instead of Unix and that fixes it.  I will submit a PR

bflad:
This appears to have been resolved awhile ago in https://github.com/hashicorp/terraform-provider-random/pull/27 👍

PR #27, opened 2018-05-22
update seed.go with UnixNano instead of Unix

wadeholler:
Fixes Issue #26


mildwonkey:
Thank you for opening both the issue and PR, @wadeholler!

PR #28, opened 2018-05-22
Hotfix: add state migration for the string resource

mildwonkey:
Set state defaults for the min_* options for users using random <= 1.3.0

### DIFF
--- a/random/resource_string_migrate.go
+++ b/random/resource_string_migrate.go
@@ -1,0 +1,46 @@
+package random
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/hashicorp/terraform/terraform"
+)
+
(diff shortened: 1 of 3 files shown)


random/resource_string_migrate.go
```diff
@@ -0,0 +1,46 @@
...
+
+func migrateStringStateV0toV1(is *terraform.InstanceState) (*terraform.InstanceState, error) {
+	if is.Empty() {
+		log.Println("[DEBUG] Empty InstanceState; nothing to migrate.")
+		return is, nil
+	}
+
+	log.Printf("[DEBUG] Attributes before migration: %#v", is.Attributes)
+
+	if v, ok := is.Attributes["min_lower"]; !ok && v == "" {
```
katbyte: I think if your trying to match either a missing or empty property this should be `!ok || v == ""`

mildwonkey:
I am going to close this and we can (please) review #29 - similar work, but by the original committer

PR #32, opened 2018-06-10
New resource random_choice

oscr:
Hi,

I've implemented a new resource called: random_choice that will randomly select one element from a list of strings provided as input. 

I realize that it's of limited use but it may be "nice to have" in certain situations where e g random_pet, random_id and random_string would work but aren't really what you're looking for.

If there are any questions please don't hesitate to ask!

Regards Oscar

### DIFF
--- a/random/resource_choice.go
+++ b/random/resource_choice.go
@@ -1,0 +1,48 @@
+package random
+
+import (
+	"errors"
+	"github.com/hashicorp/terraform/helper/schema"
+	"math/rand"
+	"time"
+)
+
+func resourceChoice() *schema.Resource {
+	return &schema.Resource{
+		Create: CreateChoice,
+		Read:   schema.Noop,
+		Delete: schema.RemoveFromState,
+
+		Schema: map[string]*schema.Schema{
+			"input": {
+				Type:     schema.TypeList,
+				Required: true,
+				ForceNew: true,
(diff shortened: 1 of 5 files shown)


random/resource_choice.go
```diff
@@ -0,0 +1,48 @@
...
+	input := d.Get("input").([]interface{})
+
+	if len(input) < 1 {
+		return errors.New("input can't be empty")
+	}
+
+	rand.Seed(time.Now().Unix())
+	result := input[rand.Intn(len(input))].(string)
+
+	d.SetId("-")
```
oscr: Question: Is this an acceptable id?

mildwonkey:
Hi @oscr, thanks for submitting this PR!

Can you please share some of your use cases for this resource? I'm curious to see where you would use it that isn't covered by other resources in `random`.  For example, you could use `random_shuffle` to shuffle a list of strings and take the first element out of the new list, or use random_integer to index into a list. 

Thanks!

oscr:
Hi @mildwonkey, 

Sorry for the late reply.

The use case is convenience. As you correctly point out it doesn't provide any unique functionality. It just makes it more convenient. I'm just offering because it was fun to write and good to have but if you feel it's doesn't add value please feel free to reject it.

PR #33, opened 2018-06-10
Check if random_shuffle is given an empty list

oscr:
Hi,

The following configuration will cause TF to get stuck in the creating step for random_shuffle

```hcl
resource "random_shuffle" "rand-shuffle" {
    input = []
} 
```

Result being:
```
random_shuffle.rand-shuffle: Creating...
  result.#: "" => "<computed>"
random_shuffle.rand-shuffle: Still creating... (10s elapsed)
...
random_shuffle.rand-shuffle: Still creating... (2m50s elapsed)
```

This change validates that at least one element is provided.

Regards Oscar

### DIFF
--- a/random/resource_shuffle.go
+++ b/random/resource_shuffle.go
@@ -1,6 +1,7 @@
 package random
 
 import (
+	"errors"
 	"github.com/hashicorp/terraform/helper/schema"
 )
 
@@ -52,6 +53,10 @@ func resourceShuffle() *schema.Resource {
 func CreateShuffle(d *schema.ResourceData, _ interface{}) error {
 	input := d.Get("input").([]interface{})
 	seed := d.Get("seed").(string)
+
+	if len(input) < 1 {
+		return errors.New("input can't be empty")
+	}
 
 	resultCount := d.Get("result_count").(int)
 	if resultCount == 0 {
--- a/random/resource_shuffle.go
+++ b/random/resource_shuffle.go
@@ -1,7 +1,6 @@
 package random
 
 import (
-	"errors"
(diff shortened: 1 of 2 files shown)


mildwonkey:
This is a great catch, @oscr.  Can you add a test case that checks for this scenario to help us avoid regression?

mildwonkey:
Hi @oscr!
I am planning on releasing this provider this week and would like it to include your PR. Can you please update it with a test? Alternatively, you could grant permission to commit to your PR branch to maintainers of this repository - Github's documentation on that feature is [here](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/), if you aren't familiar with it (I wasn't before today!)

oscr:
Hi @mildwonkey! 

Sorry for the late reply! 

If I'm not mistaken the "Allow edits from maintainers." box is already checked so you should already have the required permission. So if it's easier/faster for you to just fix it yourself please feel free to do so.

When looking at the code I started thinking that maybe responding with an error isn't the right call. Maybe it's better to respond with an empty list instead like Pythons random.shuffle() does. This is also much more straight forward to test. 

But it's your call @mildwonkey: **what do you think would be the best approach**?

mildwonkey:
Ah, sorry @oscr - the error was on my end. I'm either not actually a maintainer or (more likely) just plain doing it wrong. I'm new at this, please pardon me!

You bring up an excellent point - I also like the idea of returning an empty list. I think that's a fine idea.

oscr:
No apology required @mildwonkey. We're all just doing our best.

Well I'm glad we agree. I'll provide the update ASAP.

PR #34, opened 2018-06-10
Remove duplicate function in resource_pet

oscr:
Hi,

Removing the unneeded function ReadPet. The same functionality is already implemented in schema and I think this also makes the code clearer. 

Regards Oscar


mildwonkey:
Thank you for submitting this, @oscr!

PR #37, opened 2018-06-20
 Clarify which providers use cryptographic PRNGs in docs

zombiezen:


mildwonkey:
Thank you for working on this, @zombiezen! Good clarifications.

PR #38, opened 2018-06-30
New Resource random_uuid

straubt1:
New resource to generate a UUID rather than having to create your own concatenation. 
For some of the Azure Resources we need the ability to generate a properly formatted GUID and wanted a cleaner way to generate these ids.

### DIFF
--- a/website/docs/r/uuid.html.md
+++ b/website/docs/r/uuid.html.md
@@ -1,0 +1,53 @@
+---
+layout: "random"
+page_title: "Random: random_uuid"
+sidebar_current: "docs-random-resource-id"
+description: |-
+  Generates a random identifier.
+---
+
+# random\_id
+
+The resource `random_uuid` generates random uuid string that is intended to be
+used as unique identifiers for other resources.
+
+This resource uses the `hashicorp/go-uuid` to generate a UUID-formatted string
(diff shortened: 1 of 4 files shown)


straubt1:
@mildwonkey @katbyte Comments, questions, concerns about this change?
(Assuming you are both Owners based on recent commit history).

website/docs/r/uuid.html.md
```diff
@@ -0,0 +1,53 @@
+---
+layout: "random"
+page_title: "Random: random_uuid"
+sidebar_current: "docs-random-resource-id"
+description: |-
+  Generates a random identifier.
+---
+
+# random\_id
```
mildwonkey: I believe this should be random\_uuid

mildwonkey:
Hi @straubt1, sorry about the delay here - I'll take a look!

website/docs/r/uuid.html.md
straubt1: Pushed a change, good catch!

straubt1:
@mildwonkey Idea on next release timing?

mildwonkey:
Good question! I was just discussing this with the terraform core team earlier today. I hope to get the remaining open PRs addressed and the provider released within the next few days.

PR #39, opened 2018-07-07
added security note to random keepers

nyurik:
I saw that keepers output the content of the keepers in plain text, even if it is a secure info.  In the cases when terraform is applied via CI jobs, other people might see sensitive data.

### DIFF
--- a/website/docs/index.html.markdown
+++ b/website/docs/index.html.markdown
@@ -66,6 +66,15 @@ resource "aws_instance" "server" {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


mildwonkey:
Hi @nyurik,
Thank you for taking the time to submit this PR. I am going to decline it because the example you've included is not something we'd like users to think we are suggesting. 

However I have taken the suggestion to heart, and will open a new PR which includes a warning that `keepers` will be output in the UI as plaintext.

Thank you again!

PR #46, opened 2018-09-02
WIP: Add import support for resource_string

zimbatm:
This allows to import passwords in the terraform state instead of having them in the code.

The tests are currently failing and I don't understand why, can you help me out?

```
--- FAIL: TestAccResourceString (0.05s)
        testing.go:434: Step 1 error: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.

                (map[string]string) {
                }


                (map[string]string) (len=9) {
                 (string) (len=6) "length": (string) (len=2) "12",
                 (string) (len=5) "lower": (string) (len=4) "true",
                 (string) (len=9) "min_lower": (string) (len=1) "0",
                 (string) (len=11) "min_numeric": (string) (len=1) "0",
                 (string) (len=11) "min_special": (string) (len=1) "0",
                 (string) (len=9) "min_upper": (string) (len=1) "0",
                 (string) (len=6) "number": (string) (len=4) "true",
                 (string) (len=7) "special": (string) (len=4) "true",
                 (string) (len=5) "upper": (string) (len=4) "true"
                }

FAIL
exit status 1
FAIL    github.com/terraform-providers/terraform-provider-random/random 0.345s
```

### DIFF
--- a/random/resource_string.go
+++ b/random/resource_string.go
@@ -4,6 +4,7 @@ import (
 	"crypto/rand"
 	"math/big"
 	"sort"
+	"strconv"
 
 	"github.com/hashicorp/errwrap"
 	"github.com/hashicorp/terraform/helper/schema"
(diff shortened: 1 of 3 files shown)


zimbatm:
Closing in favour of https://github.com/tweag/terraform-provider-secret which also marks the value as sensitive by default

PR #61, opened 2019-06-06
Fix an error in website/docs/r/integer.html.md

enm10k:
When I executed a sample code, I got the following error message.

> "priority" must be in the range 1-50000

```
...
------------------------------------------------------------------------

Error: Error running plan: 1 error(s) occurred:

* aws_alb_listener_rule.main: "priority" must be in the range 1-50000 for normal rule or 99999 for default rule
```

### DIFF
--- a/website/docs/r/integer.html.md
+++ b/website/docs/r/integer.html.md
@@ -17,13 +17,13 @@ exist concurrently.
 
 ## Example Usage
 
-The following example shows how to generate a random priority between 1 and 99999 for
+The following example shows how to generate a random priority between 1 and 50000 for
 a `aws_alb_listener_rule` resource:
 
 ```hcl
 resource "random_integer" "priority" {
   min     = 1
-  max     = 99999
+  max     = 50000
   keepers = {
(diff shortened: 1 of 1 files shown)


website/docs/r/integer.html.md
```diff
@@ -23,7 +23,7 @@ a `aws_alb_listener_rule` resource:
 ```hcl
 resource "random_integer" "priority" {
   min     = 1
-  max     = 99999
+  max     = 50000
```
mildwonkey: Can you change the line above this to match the new example? 

> The following example shows how to generate a random priority between 1 and 50000 for
a `aws_alb_listener_rule` resource:

mildwonkey:
Thanks for submitting this @enm10k ! I left one comment - it would be good to update the example description to match the example.

enm10k:
@mildwonkey 
> Thanks for submitting this @enm10k ! I left one comment - it would be good to update the example description to match the example.

Thanks!
I update the example description and the import command too.

PR #65, opened 2019-07-05
clarify override_special property

benny-gold:
This is unnecessarily confusing, so I've just changed the example given.


bw-davidupton:
I just ran into this problem too.

dupton:
I also found this confusing

PR #73, opened 2019-09-12
Add a hash to random_password

tommyknows:
Hi,

## Problem

We have experienced cases where we need to hash a password with `bcrypt`. The `bcrypt` function in terraform always generates a new hash on every run (which is according to spec). However, this can be annoying to use.
For example:

```hcl
resource "random_password" "admin_password" {
  length  = 20
  special = true
}

# store the password in vault
resource "vault_generic_secret" "password" {
  path         = "/passwords/mypassword" 
  disable_read = true

  data_json = <<EOT
{
  "password": "${random_password.password.result}",
  "bcrypt_hash": "${bcrypt(random_password.password.result)}" 
}
EOT
}
```

In this example, `vault_generic_secret.password` gets changed on every run. There is also no way (at least no obvious one) to work around this problem - `ignore_changes` cannot be used when this kind of interpolation is required.

## Solution

The solution to this is to add a `bcrypt_hash` field to the `random_password` resource. This always returns the same hash (as it is stored in the terraform state), unless the password (random string) itself is changed.
With that, the above usecase can be done just by doing

```hcl
  data_json = <<EOT
{
  "password": "${random_password.password.result}",
  "bcrypt_hash": "${random_password.password.bcrypt_hash)}" 
}
EOT
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,4 +8,5 @@ require (
 	github.com/hashicorp/go-uuid v1.0.1
 	github.com/hashicorp/hcl v1.0.0 // indirect
 	github.com/hashicorp/terraform-plugin-sdk v1.11.0
+	golang.org/x/crypto v0.0.0-20200604202706-70a84ac30bf9
 )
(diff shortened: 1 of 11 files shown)


tommyknows:
@appilon 
@radeksimko

some feedback would be highly appreciated, thanks.

tommyknows:
Noticed that with the cost as a a variable, all resources would get recreated and thus may break existing things.
I have removed the cost and am just using the default value now instead.

mstarostik:
Just what I've been looking for for a while now and seems to be working as intended. Please go for it, this increases bcrypt()'s usefulness by orders of magnitude.

tommyknows:
signed the CLA, rebased and cleaned up the commit.

stefansedich:
Looking for this myself!

sorenje:
i would really like this option as well.

yunuscanemre:
hi, we are also eagerly waiting on this fix. any idea when it will be merged ?

xanderflood:
Also very eager to be able to use this! So far I haven't been able to find any way to use `bcrypt` without breaking production services, and being able to rotate password using TF only would be very helpful

xanderflood:
I posted a workaround in this [related GH issue](https://github.com/hashicorp/terraform-provider-random/issues/102#issuecomment-698605852) for anyone struggling with this

KarstenSiemer:
will this ever come? 😭

bendbennett:
Hi @tommyknows, thank you for submitting this PR and apologies that it has taken a while for us to get to this.

The proposed changes look good. There have been some significant changes to _random provider_ since your original submission. In order to incorporate your changes, could you either rebase or let us know that you'd like us to move forward with your proposal?

PR #75, opened 2019-09-20
switch to standalone tf sdk

appilon:
This PR switches to the new standalone SDK, all tests pass in TC.


appilon:
I let #76 merge instead

PR #78, opened 2019-10-09
Fix resource reference in password.html.md

HadrienPatte:


radeksimko:
Hi @HadrienPatte 

PR #79, opened 2019-10-10
replace random_string by random_password in example

yanc0:


radeksimko:
Hi @yanc0  

PR #81, opened 2019-10-11
Update password.html.md

ccampanale:
Example usage references a resource of type `random_string` but the documentation refers to the resource `random_password`; this updates the referenced resource type to properly refer to the declared `random_password` resource.


radeksimko:
Hi @ccampanale   

PR #84, opened 2019-10-21
Correction of resource key reference

hdhruna:


radeksimko:
Hi @hdhruna    

PR #86, opened 2019-10-28
Update password.html.md

h34dl355:
resource "random_password" "password"
password = random_password.password.result


radeksimko:
Hi @h34dl355     

PR #87, opened 2019-10-29
Update password.html.md. Fix typo in resource reference

LaurenceGA:
Changed `random_string` to `random_password`.
Name should be the resource name.


radeksimko:
Hi @LaurenceGA      

PR #88, opened 2019-11-01
Fix typo

hoandang:
Change random_string to random_password in example


radeksimko:
Hi @hoandang       

PR #89, opened 2019-11-05
Fix resource name in random_password example

dhoepelman:


radeksimko:
Hi @dhoepelman        

PR #92, opened 2019-11-20
Fix example code in random_password example

golightlyb:
The example code doesn't work because it uses `random_string` instead of `random_password`, giving an error:

    Error: Reference to undeclared resource
    
    A managed resource "random_string" "password" has not been declared in the root module.

https://discuss.hashicorp.com/t/random-password-example-error-what-am-i-doing-wrong/4049


radeksimko:
Hi @golightlyb          

PR #93, opened 2019-11-27
Correct a typo in using a random_password resource

stanhordiyenko:


radeksimko:
Hi @stanhordiyenko           

PR #94, opened 2019-12-02
Fix code example in random_password resource documentation

riccardomc:
The code example in the `random_password` resource uses `random_string.password.result` while it should use `random_password.password.result`.


radeksimko:
Hi @riccardomc           

PR #99, opened 2020-02-17
Enable binary acceptance test driver

kmoe:
This PR enables the new binary acceptance test driver available in v1.7.0 of the plugin SDK.

Provider acceptance tests now run real `terraform` CLI commands under the hood. The test API remains the same, and there should be no other changes to test results or output. Please see https://github.com/hashicorp/terraform-plugin-sdk/pull/262 for more details.

In addition, this PR includes the following changes:
 - Upgrade to SDK v1.7.0.
 - Force use of Go vendoring via Makefile. Prior to this, the `vendor/` directory was ignored, and Go modules were downloaded anew during each build, as can be seen in old Travis logs. Please see [this memo](https://gist.github.com/kmoe/21368bcf91cae532041921756f0a1511) for more details regarding this change.
 - Add a step to Travis CI to install Terraform prior to the build. This is currently needed by the new binary test driver; subsequent versions may install Terraform automatically.
 - Run `make testacc` during the Travis build, since this provider's acceptance tests are very quick and do not rely on any external APIs.


GNUmakefile
```diff
@@ -3,6 +3,9 @@ GOFMT_FILES?=$$(find . -name '*.go' |grep -v vendor)
 WEBSITE_REPO=github.com/hashicorp/terraform-website
 PKG_NAME=random
 
+.EXPORT_ALL_VARIABLES:
+GOFLAGS=-mod=vendor
```
appilon: I wasn't sure myself on enforcing the use of the vendor folder for local development. I would say the `vendoring` is a relic of speeding up builds for CI in a pre go cache world. I had been actually enforcing modules on just in Travis. It's something we likely want to harmonize but we can figure that out another day, I'm happy to merge anything that "just works"

.travis.yml
```diff
@@ -4,7 +4,10 @@ services:
 - docker
 language: go
 go:
-  - "1.11.x"
+  - "1.13.x"
```
appilon: Please bump the version of go in `.go-version` to match so releases and accetance tests on TeamCity remain consistent

GNUmakefile
kmoe: Yes, I agree vendoring is a bigger discussion. At the moment, most providers are using vendoring, at least on TeamCity where it is used for nightly test and release builds. I think we should try to make the development environment match the test/release environment as much as possible, which means that for the moment while vendoring is still being used in TC, it should be used for local development. Since Travis uses the Makefile, setting `GOFLAGS` here brings both local dev and PR test environments into alignment with TC.

If we remove vendoring in the future, we will have to touch either the Makefile or `.travis.yml`, wherever we have set `GOFLAGS=-mod=vendor` (the build will error if there is no `vendor/` directory, so we can't leave it in harmlessly).

What do you think?

kmoe: In fact, I think I've persuaded myself: whatever we decide to do about vendoring, it's not really related to this PR. Will remove this commit.

.travis.yml
kmoe: This didn't need to be bumped, so I've reverted it.

GNUmakefile
paddycarver: Agree with how this shook out, but wanted to know that our official internal guidance at the moment is to use vendoring when using Go Modules, and deciding against doing that should probably be an RFC.

PR #100, opened 2020-03-02
Random String and Password Import

jahantech:
This PR enables import functionality for random string. Resource state changes are used to determine if the string generation criterias were met.

### DIFF
--- a/random/resource_password.go
+++ b/random/resource_password.go
@@ -10,5 +10,8 @@ func resourcePassword() *schema.Resource {
 		Read:   readNil,
 		Delete: schema.RemoveFromState,
 		Schema: stringSchemaV1(true),
+		Importer: &schema.ResourceImporter{
+			State: ImportPassword,
+		},
 	}
 }
--- a/random/resource_pasword_test.go
+++ b/random/resource_pasword_test.go
@@ -35,6 +35,11 @@ func TestAccResourcePassword(t *testing.T) {
 					regexMatch("random_password.min", regexp.MustCompile(`([!#@])`), 1),
 				),
 			},
+			{
+				ResourceName:      "random_password.import",
+				ImportState:       true,
+				ImportStateVerify: false,
+			},
 		},
 	})
 }
--- a/random/resource_string.go
+++ b/random/resource_string.go
@@ -12,5 +12,8 @@ func resourceString() *schema.Resource {
 		MigrateState:  resourceRandomStringMigrateState,
 		SchemaVersion: 1,
 		Schema:        stringSchemaV1(false),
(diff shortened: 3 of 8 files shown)


crisp2u:
Looks good. Do you need help testing this ?

jahantech:
I have performed tests with basic as well as special characters based strings. But if you have some spare time to run through some cases that would be great. In terms of unit testing, I can add a test for  updateStringResourceAttributes method if required.

cursedcoder:
Looks good.

cursedcoder:
So far works pretty good:
```
module.tcp.random_string.db_password: Importing from ID "PgYIYHschNVy0ZeroQMTcq1mEa6WOd"...
module.tcp.random_string.db_password: Import prepared!
  Prepared random_string for import
module.tcp.random_string.db_password: Refreshing state... [id=PgYIYHschNVy0ZeroQMTcq1mEa6WOd]

Import successful!

The resources that were imported are shown above. These resources are now in
your Terraform state and will henceforth be managed by Terraform.
```

paultyng:
@jahantech can you add import steps to the acceptance testing? Here is an example in AWS: https://github.com/terraform-providers/terraform-provider-aws/blob/98b8b848ca94031b20c3e626c9d40484e3af80de/aws/resource_aws_iot_topic_rule_test.go#L33-L37

jahantech:
@paultyng  Thank you for providing the details. I have added the tests to my local branch, but they are not currently passing because my `updateStringResourceAttributes` function relies on `_, newMinClass := d.GetChange(charClassMin)` and when the tests are executed `d.GetChange(charClassMin)` always returns 0 changes. I will look into further to see how I can resolve this issue.

paultyng:
For an import, your code only runs through the import and read functions, the config file is not referenced, so you have a few options:
- make `0` work somehow
- add the additional data to the import string in addition to the password/string itself
- ignore configuration/behavioral fields in the import test
Hard for me to say which is the appropriate choice, I'd have to dig in to the functionality a little more, but wanted to give you some ideas.

jahantech:
@paultyng I would have to ignore all the fields as they are set to 0. In such case I have ImportStateVerify to false and the test now passes. Let me know if that is acceptable. Also I noticed the travis build is now failing on the master. My tests do pass locally.

paultyng:
I reworked some of the code and import testing in #104, there were some other minor issues with the repo so cleaned those up as well (vendoring). For import testing, you should explicitly ignore anything that is only user entered and not determined by the read. Its possible though that this would let people import strings that don't satisfy the create criteria, but I think that's an acceptable situation if they already have an existing password or something they want to use.

PR #113, opened 2020-04-29
update resource_password to SchemaVersion 1

jcarrothers-sap:
resource_password and resource_string share a schema definition, and thus
should share a SchemaVersion when they are using the same schema definition
function

### DIFF
--- a/random/resource_password.go
+++ b/random/resource_password.go
@@ -6,10 +6,11 @@ import (
 
 func resourcePassword() *schema.Resource {
 	return &schema.Resource{
-		Create: createStringFunc(true),
-		Read:   readNil,
-		Delete: schema.RemoveFromState,
-		Schema: stringSchemaV1(true),
+		Create:        createStringFunc(true),
+		Read:          readNil,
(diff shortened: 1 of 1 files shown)


jcarrothers-sap:
It would be appreciated if you would accept this change.  To me it seems natural that the SchemaVersion of these two resources should be aligned, given their shared code base.  Making this change would also ease a transition from an internal fork of this provider that was created prior to resource_password being accepted into a release of this provider.

bflad:
Hi @jcarrothers-sap 👋  Thank you for submitting this and sorry for the delayed review. Since this resource does not implement [state migrations](https://www.terraform.io/plugin/sdkv2/resources/state-migration), it is not necessary to increment the schema version. This resource has also now implemented the updated schema for awhile, so there is likely no practitioner benefit to introducing a similar state migration now. If we were to accept schema-based changes to this resource, that effort should likely first decouple the schemas, so this type of confusion shouldn't arise across resources. Given all this, I'm going to close this out. Thanks again for this contribution though.

PR #124, opened 2020-10-19
Test 0.14 beta

paultyng:


paultyng:
@kmoe this change should have made it run the beta, correct?

@danieldreier I think this shows green for random

PR #132, opened 2021-01-05
Fix bigint bug

kmoe:
Prior to this change, integers in state that do not fit cleanly in float64s lose their precision, leading to permadiffs when using the `random_integer` resource. As of SDK v2.4.0, setting `UseJSONNumber: true` on the resource schema changes number representation to `json.Number` throughout, preserving precision. Please see hashicorp/terraform-plugin-sdk#655 for more information.


paddycarver (approved):
Let's do it!

PR #134, opened 2021-01-24
Allow override num, lower, upper characters

sh0shin:
Allow overrides to all characters, to create passwords without ambiguous characters like `O0oIl`.


sh0shin:
Looks like Travis CI is broken

PR #136, opened 2021-01-30
feat: github actions

vfoucault:
Move CI from travis (dead long travis) to Github actions.

behavior remains the same

-> make test
-> make testacc
-> make vet.

Worth to notice that there is matrix build for terraform version, starting from (per the doc) 0.10

```
  - '0.10.8'
  - '0.11.14'
  - '0.12.30'
  - '0.13.6'
  - '0.14.5'
````


vfoucault:
Though it might work nor be relevant to test TF Version < 0.12

kmoe:
What a coincidence - we were just about to do this!

The setup is slightly different and will be tracked in https://github.com/hashicorp/terraform-provider-random/pull/140. I've added you as a co-author there.

Thanks for your contribution.

PR #158, opened 2021-04-11
accept min==max for random_integer resource

### LINKED ISSUE #157: random_integer should accept min == max
### Affected Resource(s)
- random_integer

### Terraform Configuration Files
```hcl
data "alicloud_zones" "available" {
  available_instance_type = var.instance_type
  enable_details          = true
}

resource "random_integer" "zone_index" {
  min = 0
  max = length(data.alicloud_zones.available.zones) - 1
  keepers = {
    instance_type = var.instance_type
  }
}

locals {
  zone = data.alicloud_zones.available.zones[random_integer.zone_index.result]
}
```

### Expected Behavior
Choose a random AZ that has the specified instance type available.

### Actual Behavior
When there is only one AZ that provides the specified instance type, `random_integer` refuses to generate an index of zero for me.

### References
https://github.com/hashicorp/terraform-provider-random/blob/73e21bef24618f9a90d2e9eb63c99c5da187dbf9/internal/provider/resource_integer.go#L78-L80

zhangyoufu:
fix #157

### DIFF
--- a/internal/provider/resource_integer.go
+++ b/internal/provider/resource_integer.go
@@ -77,10 +77,10 @@ func CreateInteger(_ context.Context, d *schema.ResourceData, meta interface{}) 
 	max := d.Get("max").(int)
 	seed := d.Get("seed").(string)
 
-	if max <= min {
+	if max < min {
 		return append(diags, diag.Diagnostic{
 			Severity: diag.Error,
-			Summary:  "minimum value needs to be smaller than maximum value",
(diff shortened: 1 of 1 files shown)


zhangyoufu:
@bendbennett I have explained my use case in the corresponding issue #157. In short, I tried to select a random AZ that satisfies certain conditions. But it doesn't work when there is only one AZ eligible.

PR #208, opened 2022-03-08
.github/workflows: Add 'go generate' testing for differences

bflad:
Since this project is already using `terraform-plugin-docs`, there can be differences between the source schema/markdown and generated documentation. This will ensure those differences are caught during CI.

Further documentation testing is possible via `tfproviderdocs`, however this at least covers a common issue with documentation generation.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -40,6 +40,18 @@ jobs:
       run: |
         go build -v .
 
+  generate:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/setup-go@v3
+        with:
+          go-version: '1.16'
+      - uses: actions/checkout@v3
+      - run: go generate ./...
+      - name: git diff
+        run: |
+          git diff --compact-summary --exit-code || \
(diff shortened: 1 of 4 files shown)


detro:
I might need a bit of help to understand this.

While I understand (now) how `tfproviderdocs` should be used for the doc, the test you added only uses `go generate`. And the only file that has a `go:generate` header is `main.go`. How would that work with the doc generated by `tfproviderdocs`?

This will likely be clearer to me once I setup `tfproviderdocs` against the `-tls` provider (and I added notes about this [here](https://github.com/hashicorp/terraform-providers-devex-internal/issues/94)).

detro (approved):
I left a general comment on the conversation tab, but this LGTM

bflad:
[`tfproviderdocs`](https://github.com/bflad/tfproviderdocs) != [`terraform-plugin-docs`](https://github.com/hashicorp/terraform-plugin-docs), which the latter is implemented via `go generate` comments in the code. These tools could be combined, but I was maintaining the documentation verification tool myself before the documentation generator was written. I don't know if my opinionated verification implementation is what we'd want to maintain (either in its current form or potentially added to `terraform-plugin-docs`).

PR #215, opened 2022-03-16
Feature random_id import hex chars

iamasmith:
Allows, whilst maintaining compatibility, the ability to import random_id using an alternative form to the base64 sequence. Specifically import can be directly given hex bytes as text using the form hex:a7ef6152 (equivalent of previous example p-9hUg)

### DIFF
--- a/docs/resources/id.md
+++ b/docs/resources/id.md
@@ -83,13 +83,26 @@ resource "aws_instance" "server" {
 Import is supported using the following syntax:
 
 ```shell
-# Random IDs can be imported using the b64_url with an optional prefix. This
+# Random IDs can be imported using the b64_url or a hex value with an optional prefix. This
 # can be used to replace a config value with a value interpolated from the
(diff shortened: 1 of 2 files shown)


iamasmith:
#216

iamasmith:
Noting on the subject of adding another import variant for dec:, as a decimal prefix, I refrained as this would introduce some extra expected parameterisation around the byte size of the value. Possibly another enhancement but that is beyond the scope of this use case.

iamasmith:
pushed an update to the documentation clarifying byte size when usng hex.

iamasmith:
Code updated to fix merge issues with upstream changes and adopt new error formatting

bendbennett:
@iamasmith thank you for your PR. The use-case you describe, whilst valid, does not appear to be one that is commonly requested. As a consequence we will be closing this PR but will revisit if there are further requests for the same change in the future.

PR #218, opened 2022-03-16
Remove unnecessary ID field

guineveresaenger:
The Terraform Plugin SDK already adds an ID to the schema, so this field is now unnecessary.

It causes the pulumi schema to pull a duplicate argument.

see also: https://github.com/cloudflare/terraform-provider-cloudflare/pull/1504

### DIFF
--- a/docs/resources/id.md
+++ b/docs/resources/id.md
@@ -67,6 +67,7 @@ resource "aws_instance" "server" {
 
 ### Optional
 
+- `id` (String) The ID of this resource.
 - `keepers` (Map of String) Arbitrary map of values that, when changed, will trigger recreation of resource. See [the main provider documentation](../index.html) for more information.
 - `prefix` (String) Arbitrary string to prefix the output value with. This string is supplied as-is, meaning it is not guaranteed to be URL-safe or base64 encoded.
 
@@ -76,7 +77,6 @@ resource "aws_instance" "server" {
 - `b64_url` (String) The generated id presented in base64, using the URL-friendly character set: case-sensitive letters, digits and the characters `_` and `-`.
 - `dec` (String) The generated id presented in non-padded decimal digits.
 - `hex` (String) The generated id presented in padded hexadecimal digits. This result will always be twice as long as the requested byte length.
-- `id` (String) The generated id presented in base64 without additional transformations or prefix.
 
 ## Import
 
--- a/internal/provider/resource_id.go
+++ b/internal/provider/resource_id.go
@@ -85,12 +85,6 @@ exist concurrently.
 				Type:        schema.TypeString,
 				Computed:    true,
 			},
-
-			"id": {
-				Description: "The generated id presented in base64 without additional transformations or prefix.",
-				Type:        schema.TypeString,
-				Computed:    true,
-			},
 		},
 	}
 }


bflad:
Hi @guineveresaenger 👋  Thank you for submitting this.

While I'm not intimately familiar with `pulumi schema`, the canonical data source for a Terraform Provider schema information in external programs is currently the [`terraform providers schema -json` command](https://www.terraform.io/cli/commands/providers/schema). This is how tooling such as [`terraform-ls` (Terraform Language Server)](https://github.com/hashicorp/terraform-ls) operates. If the Pulumi tooling is importing or transliterating the code of this project, we do not support that type of activity. As a best guess in that case, it would need to account for this type of schema information potentially being present in the code, filtering out the (non-existent) duplicate.

To get the schema information via the Terraform CLI command, a placeholder configuration can be used, such as:

```terraform
terraform {
  required_providers {
    random = {
      source  = "hashicorp/random"
    }
  }
}
```

Then the following commands executed:

```shell
terraform init
terraform providers schema -json
```

For example, `terraform providers schema -json | jq '.provider_schemas["registry.terraform.io/hashicorp/random"].resource_schemas["random_id"]'` returns:

```json
{
  "version": 0,
  "block": {
    "attributes": {
      "b64_std": {
        "type": "string",
        "description": "The generated id presented in base64 without additional transformations.",
        "description_kind": "markdown",
        "computed": true
      },
      "b64_url": {
        "type": "string",
        "description": "The generated id presented in base64, using the URL-friendly character set: case-sensitive letters, digits and the characters `_` and `-`.",
        "description_kind": "markdown",
        "computed": true
      },
      "byte_length": {
        "type": "number",
        "description": "The number of random bytes to produce. The minimum value is 1, which produces eight bits of randomness.",
        "description_kind": "markdown",
        "required": true
      },
      "dec": {
        "type": "string",
        "description": "The generated id presented in non-padded decimal digits.",
        "description_kind": "markdown",
        "computed": true
      },
      "hex": {
        "type": "string",
        "description": "The generated id presented in padded hexadecimal digits. This result will always be twice as long as the requested byte length.",
        "description_kind": "markdown",
        "computed": true
      },
      "id": {
        "type": "string",
        "description": "The generated id presented in base64 without additional transformations or prefix.",
        "description_kind": "markdown",
        "computed": true
      },
      "keepers": {
        "type": [
          "map",
          "string"
        ],
        "description": "Arbitrary map of values that, when changed, will trigger recreation of resource. See [the main provider documentation](../index.html) for more information.",
        "description_kind": "markdown",
        "optional": true
      },
      "prefix": {
        "type": "string",
        "description": "Arbitrary string to prefix the output value with. This string is supplied as-is, meaning it is not guaranteed to be URL-safe or base64 encoded.",
        "description_kind": "markdown",
        "optional": true
      }
    },
    "description": "\nThe resource `random_id` generates random numbers that are intended to be\nused as unique identifiers for other resources.\n\nThis resource *does* use a cryptographic random number generator in order\nto minimize the chance of collisions, making the results of this resource\nwhen a 16-byte identifier is requested of equivalent uniqueness to a\ntype-4 UUID.\n\nThis resource can be used in conjunction with resources that have\nthe `create_before_destroy` lifecycle flag set to avoid conflicts with\nunique names during the brief period where both the old and new resources\nexist concurrently.\n",
    "description_kind": "markdown"
  }
}
```

In addition with this change, the description information is intentionally defined as part of the schema to provide practitioners with details about the expected contents of the attribute value, which is output into the automatically generated documentation for the resource. Removing it means the provider documentation is less useful and we could never provide any guarantees that it would not be reintroduced, since it is generally useful for practitioners. We anticipate this type of schema description information to become more prevalent in other providers (especially for `id` attributes) as they are converted over to the documentation generator.

Given all this, I'm going to close this out since its not a change we would be willing to accept. Thanks again for raising this though.

stack72:
Just to add a little extra here - the addition of an Id field is superfluous. I understand that it's specific to the Pulumi schema but it's not needed as it's part of d.setId and is actually a Hangup of the change to that ecosystem

As a former member of the tf ecosystem, I don't see why this PR wouldn't be merged - I can fork the provider and use that.

I thank you for the time you gave here @bflad

Issue #230, opened 2022-04-12
Replace usage of deprecated fields in schema.Resource


bendbennett:
### Terraform CLI and Provider Versions

Terraform v1.1.7
random provider v3.1.2

### Use Cases or Problem Statement

The random provider is using deprecated fields within schema.Resource.
For example [resource_id](https://github.com/hashicorp/terraform-provider-random/blob/v3.1.2/internal/provider/resource_id.go#L31) uses the `Create` field which has been deprecated in favour of `CreateContext` or `CreateWithoutTimeout`.

### Proposal

Replace usage of deprecated fields in schema.Resource (e.g., replace `Create` with `CreateContext` or `CreateWithoutTimeout`). 

### How much impact is this issue causing?

Low

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #231, opened 2022-04-12
Replace usage of deprecated fields in schema.Resource

bendbennett:
Closes https://github.com/hashicorp/terraform-provider-random/issues/230.


internal/provider/resource_integer.go
```diff
@@ -17,11 +19,11 @@ func resourceInteger() *schema.Resource {
 			"This resource can be used in conjunction with resources that have the `create_before_destroy` " +
 			"lifecycle flag set, to avoid conflicts with unique names during the brief period where both the " +
 			"old and new resources exist concurrently.",
-		Create: CreateInteger,
-		Read:   schema.Noop,
-		Delete: schema.RemoveFromState,
+		CreateContext: CreateInteger,
+		ReadContext:   schema.NoopContext,
+		DeleteContext: DeleteInteger,
```
detro, line 24: It's annoying that there is no `schema.RemoveFromStateContext`

internal/provider/resource_pet.go
```diff
@@ -22,9 +23,9 @@ func resourcePet() *schema.Resource {
 			"This resource can be used in conjunction with resources that have the `create_before_destroy` " +
 			"lifecycle flag set, to avoid conflicts with unique names during the brief period where both the old " +
 			"and new resources exist concurrently.",
-		Create: CreatePet,
-		Read:   schema.Noop,
-		Delete: schema.RemoveFromState,
+		CreateContext: CreatePet,
+		ReadContext:   schema.NoopContext,
+		DeleteContext: DeletePet,
```
detro, line 28: as per above

detro (approved):
A suggestion, but otherwise it looks good.

internal/provider/resource_integer.go
bendbennett: @detro your suggestion makes sense to me. Have created a shared `DeleteContext` func and added it to `provider.go`.

internal/provider/resource_pet.go
bendbennett: Done (see above).

internal/provider/resource.go
```diff
@@ -0,0 +1,13 @@
+package provider
+
+import (
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func DeleteContext(_ context.Context, d *schema.ResourceData, _ interface{}) diag.Diagnostics {
```
detro: Nitpick: would it make sense to use a name for this that doesn't clash with the field it's used in?

Reading below:

```golang
{
  // ...
  DeleteContext: DeleteContext,
  // ...
}
```

it's a bit ... odd.

Admittedly, coming up with a name can be hard (as usually, naming things always is).
How about `removeResourceFromState()`, since that's what `d.SetId("")` is for?

detro (approved):

PR #246, opened 2022-04-26
Adopt tflog

### LINKED ISSUE #245: Adopt tflog

Terraform v1.1.8
Provider 3.1.3

Currently, the provider generates logs by using the [log package](https://pkg.go.dev/log) within the standard library.

### Proposal

Adopt [tflog](https://www.terraform.io/plugin/log/writing) package.

Low

### Additional Information

_No response_

### Code of Conduct

bendbennett:
Closes #245

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,9 +4,9 @@ go 1.17
 
 require (
 	github.com/dustinkirkland/golang-petname v0.0.0-20191129215211-8e5a1ed0cff0
-	github.com/hashicorp/errwrap v1.1.0
 	github.com/hashicorp/go-uuid v1.0.3
 	github.com/hashicorp/terraform-plugin-docs v0.7.0
+	github.com/hashicorp/terraform-plugin-log v0.3.0
 	github.com/hashicorp/terraform-plugin-sdk/v2 v2.14.0
 )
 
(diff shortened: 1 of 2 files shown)


bendbennett:
@bflad given https://github.com/hashicorp/terraform-plugin-log/issues/52, I'm assuming the changes in the PR will not yield the desired results as `context.TODO()` is being used. Can you suggest how logging should be handled as the [SDK documentation](https://github.com/hashicorp/terraform-plugin-sdk/blob/main/helper/schema/resource.go#L91) indicates that `MigrateState` should remain in place for compatibility with existing state?

detro:
I might be wrong, but I think you essentially don't have to setup `context.Context` ever: as long as you use the interfaces and types that provide you with a context (and yes, will need to move from `MigrateState` to `StateUpgraders`), you can then pass that context to `tflog.*` calls.

Additionally, I'm planning to make use of the "subsystem facility" like here https://pkg.go.dev/github.com/hashicorp/terraform-plugin-log/tflog#SubsystemDebug - for example, each resource could be considered a subsystem.

detro:
Ah crap :( 

Then let's see what @bflad thinks about the issue you highlighted above.

PR #247, opened 2022-04-27
Clean-up Tests

bendbennett:
### Objectives
- Remove obsolete references to `b64` in `resource_id_test` as `b64` was [deprecated and removed since v3.0.0](https://github.com/hashicorp/terraform-provider-random/issues/10).
- Add explanatory comments to `resource_password_test` to clarify why `Import` had to be run in a separate test.
- Separate `resources` into separate configs.

### DIFF
--- a/internal/provider/resource_id_test.go
+++ b/internal/provider/resource_id_test.go
@@ -27,17 +27,31 @@ func TestAccResourceID(t *testing.T) {
 						b64StdLen: 8,
 						hexLen:    8,
 					}),
-					testAccResourceIDCheck("random_id.bar", &idLens{
-						b64UrlLen: 12,
-						b64StdLen: 14,
-						hexLen:    14,
-					}),
 				),
 			},
 			{
 				ResourceName:      "random_id.foo",
 				ImportState:       true,
 				ImportStateVerify: true,
+			},
+		},
+	})
+}
+
+func TestAccResourceID_importWithPrefix(t *testing.T) {
+	resource.UnitTest(t, resource.TestCase{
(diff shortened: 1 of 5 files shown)


internal/provider/resource_id_test.go
```diff
@@ -24,24 +23,36 @@ func TestAccResourceID(t *testing.T) {
...
+	})
+}
+
+func TestAccResourceID_importWithPrefix(t *testing.T) {
+	resource.UnitTest(t, resource.TestCase{
+		PreCheck:          func() { testAccPreCheck(t) },
+		ProviderFactories: testAccProviders,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccResourceIDConfigWithPrefix,
```
detro, line 47: Would it make sense to move all those strings inline?

I feel like our tests are already an articulate structure, as `resource.Test()` require it. Whenever possible, having `Config` string defined there and then, imho, makes for easier reading (i.e. I don't have to scroll up and down to find what I'm testing and see how I'm testing it).

The fact that you are already splitting those strings into separate resources does help a lot already.

detro (approved):
I don't know your preference, but I'd like to propose that the configuration code is put inline to the `Config:` field instead of constants. Especially with you now starting to split some of them.

What do you think? It's ok if we don't do this though, so I'm going to approve as everything else LGTM.

internal/provider/resource_id_test.go
bendbennett: As some of the same `Config` strings are used in multiple locations I think I'm going to leave these as is for now. Happy to revisit at a later date if it seems like inlining would be better.

PR #249, opened 2022-05-03
Migrate from SDKv2 => Framework

### LINKED ISSUE #177: Migrate to terraform-plugin-framework
Migrate from https://github.com/hashicorp/terraform-plugin-sdk to https://github.com/hashicorp/terraform-plugin-framework.

bendbennett:
### Objectives
- Migrate provider from [SDKv2](https://github.com/hashicorp/terraform-plugin-sdk) to [Framework](https://github.com/hashicorp/terraform-plugin-framework) to align with [RFC TF-483 Utility Provider Upgrade](https://docs.google.com/document/d/1_zUbW-PYvsr0IKoO22GhANcmiaL9OxAwYCLkHomL2-0/edit#).
- ~~Use [muxing](https://github.com/hashicorp/terraform-plugin-mux) so the changes will be compatible with `tf5` protocol and TF >= 0.12.~~
- Replicate the current behaviour of the provider (i.e., ~~[v3.1.3](https://github.com/hashicorp/terraform-provider-random/releases/tag/v3.1.3)~~ [v3.2.0](https://github.com/hashicorp/terraform-provider-random/releases/tag/v3.2.0)).

### Items Requiring Resolving
- ~~Currently, there is a [failing test](https://github.com/hashicorp/terraform-provider-random/runs/6324680529?check_suite_focus=true) which is the result of a [rounding issue](hashicorp/terraform-plugin-go#189) that is currently under investigation. This should be fixed by [terraform-plugin-go/pull/190](https://github.com/hashicorp/terraform-plugin-go/pull/190).~~ This has been fixed in terraform-plugin-go and `v0.9.1` is now being used.

### Breaking Changes
As described in the CHANGELOG:

```
NOTES:

* Provider has been re-written using the new [`terraform-plugin-framework`](https://www.terraform.io/plugin/framework) ([#177](https://github.com/hashicorp/terraform-provider-random/pull/177)).

BREAKING CHANGES:

* [Terraform `>=1.0`](https://www.terraform.io/language/upgrade-guides/1-0) is now required to use this provider.

* resource/random_password: Deprecated attribute `number` has been removed ([266](https://github.com/hashicorp/terraform-provider-random/issues/266)).
* resource/random_string: Deprecated attribute `number` has been removed ([266](https://github.com/hashicorp/terraform-provider-random/issues/266)). 
```

Closes #177


internal/provider_fm/resource_id.go
```diff
@@ -0,0 +1,219 @@
...
+`,
+		Attributes: map[string]tfsdk.Attribute{
+			"keepers": {
+				Description: "Arbitrary map of values that, when changed, will trigger recreation of " +
+					"resource. See [the main provider documentation](../index.html) for more information.",
+				Type: types.MapType{
+					ElemType: types.StringType,
+				},
+				Optional:      true,
+				PlanModifiers: []tfsdk.AttributePlanModifier{tfsdk.RequiresReplace()},
```
detro, line 42: Ah, this is how `ForceNew` is done - correct?

internal/provider_fm/resource_id.go
```diff
@@ -0,0 +1,219 @@
...
+					"always be twice as long as the requested byte length.",
+				Type:     types.StringType,
+				Computed: true,
+			},
+			"dec": {
+				Description: "The generated id presented in non-padded decimal digits.",
+				Type:        types.StringType,
+				Computed:    true,
+			},
+			"id": {
```
detro, line 80: Once `v0.8.0` is shipped (https://github.com/hashicorp/terraform-plugin-docs/pull/134), the docs tools won't need us to set explicitly the `id` in the schema anymore (but it will still be possible).

internal/provider_fm/resource_id.go
```diff
@@ -0,0 +1,219 @@
...
+		p: *(p.(*provider)),
+	}, nil
+}
+
+type resourceID struct {
+	p provider
+}
+
+func (r resourceID) Create(ctx context.Context, req tfsdk.CreateResourceRequest, resp *tfsdk.CreateResourceResponse) {
+	if !r.p.configured {
```
detro, line 100: I know you saw this in the scaffolding too, but I wonder why do we need this pattern: don't we have guarantees that a callstack, reaching this method, definitely has `p.configured` true?

internal/provider_fm/resource_id.go
bendbennett: Nice! Will look out for that.

internal/provider_fm/resource_id.go
bendbennett: That's a good question. I'm not sure. Perhaps @bflad can steer us in the right direction?

bflad: While the `ConfigureProvider` RPC should always occur before `ApplyResourceChange` RPCs, it is possible that the provider configuration contains unknown values during the `ConfigureProvider` call. For example, a configuration may create an AWS RDS instance resource, then immediately reference attribute values of that resource in the provider configuration for say, a MySQL or Postgres provider that then has its own resources. Terraform CLI will typically perform provider configurations early in the graph for various reasons, not necessarily slotting in the provider configuration after the dependency resource creation.

Most providers have opted for the simpler setup of complaining during provider configuration if there are unknown values, rather than attempting to defer provider configuration logic to the last possible moment. There is limitations in that deferral situation, since the provider configuration is not present in other RPCs. I'm guessing most, if not all, of the utility providers can do the same and skip the configuration check during CRUD, since they typically don't have provider configuration at all anyways. Of course, this can be changed in the future if necessary.

Reference: https://github.com/hashicorp/terraform/issues/4149
Reference: https://github.com/hashicorp/terraform/issues/30937

internal/provider/string.go
```diff
@@ -1,219 +1,382 @@
...
+func (m minIntValidator) Validate(ctx context.Context, req tfsdk.ValidateAttributeRequest, resp *tfsdk.ValidateAttributeResponse) {
+	t := req.AttributeConfig.(types.Int64)
+
+	if t.Value < m.val {
+		resp.Diagnostics.AddError(
+			fmt.Sprintf("expected attribute to be at least %d, got %d", m.val, t.Value),
+			fmt.Sprintf("expected attribute to be at least %d, got %d", m.val, t.Value),
 		)
+	}
+}
```
detro, line 182: Nitpick: Would it make sense to have the Validators and PlanModifiers into their own files?
Like a `validators.go` and a `plan_modifiers.go`?

I might be wrong but they feel a bit "out of their context" in here, given that some are also used in other contexts (not just `random_string` and `random_password`).

main.go
```diff
@@ -1,9 +1,15 @@
...
-
-	"github.com/terraform-providers/terraform-provider-random/internal/provider"
+	"context"
+	"github.com/hashicorp/terraform-plugin-framework/providerserver"
+	"github.com/hashicorp/terraform-plugin-go/tfprotov5"
+	"github.com/hashicorp/terraform-plugin-go/tfprotov5/tf5server"
+	"github.com/hashicorp/terraform-plugin-go/tfprotov6"
+	"github.com/hashicorp/terraform-plugin-mux/tf5muxserver"
+	"github.com/hashicorp/terraform-plugin-mux/tf6to5server"
+	"github.com/terraform-providers/terraform-provider-random/internal/provider_fm"
```
detro, line 11: I think this shouldn't compile: from what I see `provider_fm` was just a temporary module that you have got rid of, now that the PR is ready.

detro (commented):
Left a few comments and nitpicks. In general this is looking really good.

Aside from those minor things, I do have a question (I might have asked this before already, so I apologise if I did): what's the plan in terms of Protocol 5 -> 6 support?

As part of the "Utility Providers RFC" we established we would be moving to 6 also as a way to "nudge" the ecosystem.

Is the intention here to do one last Protocol 5 release and then upgrade to 6, by removing mux? Or do you have a different plan?

internal/provider/resource_id.go
```diff
@@ -29,152 +31,182 @@ the ` + "`create_before_destroy`" + ` lifecycle flag set to avoid conflicts with
...
+	byteLength := plan.ByteLength.Value
+	bytes := make([]byte, byteLength)
 
-	bytes, err := base64.RawURLEncoding.DecodeString(base64Str)
+	n, err := rand.Reader.Read(bytes)
+	if int64(n) != byteLength {
+		resp.Diagnostics.AddError(
+			"generated insufficient random bytes: %s",
+			fmt.Sprintf("generated insufficient random bytes: %s", err),
+		)
```
bflad, line 116: Nit: Diagnostic summaries should generally be a "title" of sorts, e.g. what's the higher level issue?, leaving the details to explain the actual problem, offer troubleshooting advice, and/or next steps. Similar comment applies to any other place that's copying contents between summary and details.

For something like this, it'd be great to say something like:

```suggestion
		resp.Diagnostics.AddError(
			"Randomness Generation Error",
			"While attempting to generate a random value for this resource, an insufficient number of random bytes were generated. Most commonly, this is a hardware or operating system issue where their random number generator does not provide a sufficient randomness source. Otherwise, it may represent an issue in the randomness handling of the provider.\n\n"+
				"Retry the Terraform operation. If the error still occurs or happens regularly, please contact the provider developer with hardware and operating system information.\n\n"+
				fmt.Sprintf("Original Error: %s", err)
		)
```

If desired, this diagnostic can then be wrapped in a quick function and reused for other resources.

Reference: https://www.terraform.io/plugin/framework/diagnostics

internal/provider/resource_id.go
```diff
@@ -4,17 +4,19 @@ import (
 	"context"
 	"crypto/rand"
 	"encoding/base64"
-	"encoding/hex"
+	hex2 "encoding/hex"
```
bflad, line 7: In general with Go, it is better to avoid renaming imports, especially those from standard library. Instead I'd recommend renaming the variable that's trying to use `hex` instead. 👍 

Reference: https://github.com/golang/go/wiki/CodeReviewComments#imports

internal/provider/resource_integer.go
```diff
@@ -3,138 +3,184 @@ package provider
 import (
 	"context"
 	"fmt"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
```
bflad, line 6: Super nit: Typically imports outside the standard library should be in a separate import group like it was previously. Tooling like `goimports` will automatically do this.

Reference: https://github.com/golang/go/wiki/CodeReviewComments#imports

internal/provider/string.go
bflad: Personally, I agree with the above. It'll also help us as we consider making a corpus of "common" validators in the near future.

internal/provider/string.go
```diff
@@ -1,219 +1,382 @@
...
+func (m minIntValidator) Description(context.Context) string {
+	return "MinInt validator ensures that attribute is at least val"
+}
+
+func (m minIntValidator) MarkdownDescription(context.Context) string {
+	return "MinInt validator ensures that attribute is at least `val`"
+}
+
+func (m minIntValidator) Validate(ctx context.Context, req tfsdk.ValidateAttributeRequest, resp *tfsdk.ValidateAttributeResponse) {
+	t := req.AttributeConfig.(types.Int64)
```
bflad, line 174: I'd strongly encourage avoiding a direct type assertion in validators. This can cause a panic if it receives an unexpected type. My recommendation would be to at least use the two-value form of type assertion to return an error on an unexpected type and return early, e.g.

```go
t, ok := req.AttributeConfig.(types.Int64)

if !ok {
  resp.Diagnostics.AddError()
  return
}
```

Or use something similar to the terraform.io documentation here with `tfsdk.ValueAs()`:

```go
var t types.Int64
diags := tfsdk.ValueAs(ctx, req.AttributeConfig, &t)
resp.Diagnostics.Append(diags...)

if diags.HasError() {
  return
}
```

Reference: https://go.dev/doc/effective_go#interface_conversions
Reference: https://www.terraform.io/plugin/framework/validation#creating-attribute-validators

bflad (commented):
I'd also like to discuss the merits of trying to do this in a minor release versus major release, since the migration testing is currently manual and there could be unexpected value behaviors when upgrading.

.github/workflows/test.yml
```diff
@@ -34,6 +34,7 @@ jobs:
       uses: golangci/golangci-lint-action@v3
       with:
         version: latest
+        skip-cache: true
```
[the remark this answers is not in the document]
bendbennett: Removed the `skip-cache` as unnecessary.

internal/provider/provider.go
```diff
@@ -2,32 +2,38 @@ package provider
...
-	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
 )
 
-func init() {
-	schema.DescriptionKind = schema.StringMarkdown
+type provider struct {
+	configured bool
```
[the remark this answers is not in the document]
bendbennett: Agreed. Have removed.

internal/provider/resource_id.go
```diff
@@ -29,152 +31,182 @@ the ` + "`create_before_destroy`" + ` lifecycle flag set to avoid conflicts with
...
 }
 
-func ImportID(_ context.Context, d *schema.ResourceData, _ interface{}) ([]*schema.ResourceData, error) {
-	id := d.Id()
+func (r resourceID) Update(ctx context.Context, req tfsdk.UpdateResourceRequest, resp *tfsdk.UpdateResourceResponse) {
+	// Intentionally left blank.
+}
+
+func (r resourceID) Delete(ctx context.Context, req tfsdk.DeleteResourceRequest, resp *tfsdk.DeleteResourceResponse) {
+	resp.State.RemoveResource(ctx)
```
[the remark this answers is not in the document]
bendbennett: Thanks for the heads-up. I was not aware of that. Have removed all explicit calls to `resp.State.RemoveResource`.

internal/provider/resource_id.go
```diff
@@ -29,152 +31,182 @@ the ` + "`create_before_destroy`" + ` lifecycle flag set to avoid conflicts with
...
+	diags = resp.State.Set(ctx, i)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
 	}
+}
 
-	return nil
+func (r resourceID) Read(ctx context.Context, req tfsdk.ReadResourceRequest, resp *tfsdk.ReadResourceResponse) {
+	// Intentionally left blank.
```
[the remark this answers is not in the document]
bendbennett: Good call. Have added more explicit comments for `Read` and `Update` operations.

internal/provider/string.go
bendbennett: Makes sense to me. Have moved `validators` and `plan modifiers` to separate files.

main.go
bendbennett: You are right sir. Fixed

internal/provider/resource_id.go
bendbennett: I've used your wording and expanded the summary and detail messages for other diagnostics errors.

internal/provider/resource_id.go
bendbennett: Ooops. My bad. Fixed.

internal/provider/resource_integer.go
bendbennett: Looks like I forgot to configure `goimports` in GoLand. Fixed

bendbennett:
> Left a few comments and nitpicks. In general this is looking really good.
> 
> Aside from those minor things, I do have a question (I might have asked this before already, so I apologise if I did): what's the plan in terms of Protocol 5 -> 6 support?
> 
> As part of the "Utility Providers RFC" we established we would be moving to 6 also as a way to "nudge" the ecosystem.
> 
> Is the intention here to do one last Protocol 5 release and then upgrade to 6, by removing mux? Or do you have a different plan?

TBH the mux is just present as I was using this whilst migrating one resource at a time from SDKv2 to Framework. As @bflad indicates, probably best to remove the muxing and do a breaking change release.

bendbennett:
> I'd also like to discuss the merits of trying to do this in a minor release versus major release, since the migration testing is currently manual and there could be unexpected value behaviors when upgrading.

Agree, that it's probably best to remove the muxing and have a breaking change major release.

bendbennett:
I have separated out the resources into their own packages as it seems like there are details that are specific to the resources that shouldn’t be exposed, for instance, `models` and unexported funcs (e.g., `upgradePasswordStateV1toV2()`, `generateHash()` in password), but I'm happy to refactor if people would rather we kept the resources with the provider.

main.go
```diff
@@ -17,5 +20,10 @@ import (
 //go:generate go run github.com/hashicorp/terraform-plugin-docs/cmd/tfplugindocs
 
 func main() {
-	plugin.Serve(&plugin.ServeOpts{ProviderFunc: provider.New})
+	err := providerserver.Serve(context.Background(), provider.New, providerserver.ServeOpts{
```
detro, line 23: Would you mind adding the `-debug` flag here, similarly to https://github.com/hashicorp/terraform-provider-tls/blob/main/main.go#L16 ?

This is not strictly in topic with this PR, but it seems small enough that I thought I'd mention it.

detro (approved):
A suggestion and a nitpick/question about the code restructuring.

internal/provider/provider_resource_id_test.go
```diff
@@ -0,0 +1,98 @@
+package provider
```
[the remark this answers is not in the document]
bendbennett: I have separated out the resources into their own packages as it seems like there are details that are specific to the resources that shouldn’t be exposed, for instance, models and unexported funcs (e.g., `upgradePasswordStateV1toV2()`, `generateHash()` in password), but I'm happy to refactor if people would rather we kept the resources with the provider.

In terms of the tests, I must admit I don't really see them as "resource" tests, we're exercising the provider and a resource it exposes (e.g., password, string etc) through the interaction with TF core , so I see them as provider tests. The tests that are specific to resources themselves (e.g., `TestUpgradePasswordStateV0toV2`) still live with the resource code, but there's an argument for removing these sorts of tests as they're essentially testing "private" (unexported) functions which is generally frowned upon.

Another, indirectly related aspect of the refactoring has to do with import cycles. Perhaps one of the reasons a provider and its resources normally reside within the same package (e.g., `package provider`) is to avoid the issue of import cycles when a resource needs to use the API client that is typically configured in the `Configure()` func, for instance:

```go
func (p *provider) Configure(ctx context.Context, req tfsdk.ConfigureProviderRequest, resp *tfsdk.ConfigureProviderResponse) {
    p.client = client.NewClient()
}
```

Access to the `client` is then [usually achieved in the resource by using something like the following](https://learn.hashicorp.com/tutorials/terraform/plugin-framework-create?in=terraform/providers#explore-resource-order) in the `NewResource()` func:

```go
func (r resource) NewResource(_ context.Context, p tfsdk.Provider) (tfsdk.Resource, diag.Diagnostics) {
    return resource{
        p: *(p.(*provider)),
    }, nil
}
```

This seems a slightly peculiar pattern to me, as we have a `provider` that has `resources` but each resource also contains the provider and accesses "private" (unexported) fields on the provider in order to get the `client` for instance. 

Although a client is not being used by the random provider, I'd be keen to encourage a pattern whereby the client can be obtained from the provider by checking to see if it implements an interface that allows access to the client without exposing any of the underlying "private" (unexported) fields of the provider. Perhaps something along the following lines:

```go
package client

type Client interface {
     Client() API
}

type API interface {
     GetString() string
}

func NewClient() *client {
    return &client{
        msg: "hello",
    }
}

type client struct {
    msg string
}

func (c *client) GetString() string {
     return c.msg
}
```

```go
package provider

func NewProvider() tfsdk.Provider {
    return &provider{}
}

var _ tfsdk.Provider = (*provider)(nil)

type provider struct {
    client client.API
}

func (p *provider) Client() client.API {
    return p.client
}

func (p *provider) Configure(ctx context.Context, req tfsdk.ConfigureProviderRequest, resp *tfsdk.ConfigureProviderResponse) {
    p.client = client.NewClient()
}
```

```go
package resource

func (r *resourceType) NewResource(_ context.Context, p tfsdk.Provider) (tfsdk.Resource, diag.Diagnostics) {
    c, ok := p.(client.Client)
    if !ok {
        panic("not ok")
    }

    return &resource{
        client: c.Client(),
    }, nil
}

var (
_ tfsdk.Resource                = (*resource)(nil)
_ tfsdk.ResourceWithImportState = (*resource)(nil)
)

type resource struct {
    client client.API
}

func (r *resource) Create(ctx context.Context, req tfsdk.CreateResourceRequest, resp *tfsdk.CreateResourceResponse) {
    fmt.Println(r.client.GetString())
``` 

So this is a very long-winded way of saying that the refactoring also encourages decoupling the provider and resources.

main.go
bendbennett: Good shout, have added.

internal/provider/provider_resource_id_test.go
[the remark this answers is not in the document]
detro: 👍 I can't argue against incapsulation: I frankly dislike how Go doesn't support controlling visibility more granularly within the same package, so I guess this is the most idiomatic way to go.

In the future I might copy this approach.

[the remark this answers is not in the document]
bflad: Hey folks 👋 I'd like to chat about this more, but it is probably easier to do over a higher bandwidth medium.

In Go terms, the only "publicly facing" things in a provider codebase would be anything outside the `internal/` directory. As long as things abide by that, you can export all types, methods, etc. without suggesting any sort of exported API that needs to worry about compatibility. Historically, most small providers are implemented in a "flat" package structure, e.g. everything terraform-plugin-sdk/framework related in `internal/provider`, to make things easier for contributors and avoid overcomplicating things with extra interfaces, etc.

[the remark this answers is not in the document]
bflad: That being said, providers typically avoid this by putting the shared client type in another package so both the provider code and data source/resource code can all reference it without an import cycle.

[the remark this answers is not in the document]
bendbennett: @bflad I've gone ahead and refactored to have all of the resources and provider in `package provider` as I think the case for making things easier for contributors is a fair one. If you can take another look and let me know what you think. Thanks.

internal/provider/resource_integer_test.go
```diff
@@ -5,19 +5,21 @@ import (
...
 		Steps: []resource.TestStep{
 			{
-				Config: testRandomIntegerBasic,
+				Config: `resource "random_integer" "integer_1" {
+   							min  = 1
+							max  = 3
+   							seed = "12345"
+						}`,
 				Check: resource.ComposeTestCheckFunc(
-					testAccResourceIntegerBasic("random_integer.integer_1"),
```
bflad, line 20: Nice work on removing these 👍

bflad (approved):
Looks good to me 🚀 Excellent work

internal/provider/resource_id.go
```diff
@@ -29,152 +36,204 @@ the ` + "`create_before_destroy`" + ` lifecycle flag set to avoid conflicts with
...
 
-	bytes, err := base64.RawURLEncoding.DecodeString(base64Str)
+	n, err := rand.Reader.Read(bytes)
+	if int64(n) != byteLength {
+		resp.Diagnostics.AddError(
+			"Randomness Generation Error",
+			"While attempting to generate a random value for this resource, an insufficient number of random bytes were generated. Most commonly, this is a hardware or operating system issue where their random number generator does not provide a sufficient randomness source. Otherwise, it may represent an issue in the randomness handling of the provider.\n\n"+
+				"Retry the Terraform operation. If the error still occurs or happens regularly, please contact the provider developer with hardware and operating system information.\n\n"+
+				fmt.Sprintf("Original Error: %s", err),
+		)
```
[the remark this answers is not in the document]
bendbennett: 👍 moved.

PR #254, opened 2022-05-13
Add a hash to random_password with StateUpgrader

bendbennett:
This is evaluating whether the changes proposed in https://github.com/hashicorp/terraform-provider-random/pull/73 can be combined with a StateUpgrader to avoid the situation where including the `bcrypt_hash` field generates an empty value in state on first use.

### DIFF
--- a/internal/provider/resource_password.go
+++ b/internal/provider/resource_password.go
@@ -1,22 +1,92 @@
 package provider
 
 import (
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"golang.org/x/crypto/bcrypt"
 )
 
 func resourcePassword() *schema.Resource {
+	passwordSchema := stringSchemaV1(true)
+	passwordSchema["bcrypt_hash"] = &schema.Schema{
+		Type:      schema.TypeString,
+		Computed:  true,
+		Sensitive: true,
+	}
+
+	create := func(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+		diags := createStringFunc(true)(ctx, d, meta)
+		if diags.HasError() {
+			return diags
+		}
+
+		if d.Get("bcrypt_hash") == "" {
+			hash, err := generateHash(d.Get("result").(string))
+			if err != nil {
+				diags = append(diags, diag.Errorf("err: %s", err)...)
+				return diags
+			}
+
+			if err := d.Set("bcrypt_hash", hash); err != nil {
+				diags = append(diags, diag.Errorf("err: %s", err)...)
+				return diags
+			}
+		}
+
+		return nil
+	}
+
 	return &schema.Resource{
 		Description: "Identical to [random_string](string.html) with the exception that the result is " +
 			"treated as sensitive and, thus, _not_ displayed in console output. Read more about sensitive " +
 			"data handling in the [Terraform documentation](https://www.terraform.io/docs/language/state/sensitive-data.html).\n" +
 			"\n" +
 			"This resource *does* use a cryptographic random number generator.",
-		CreateContext: createStringFunc(true),
+		CreateContext: create,
 		ReadContext:   readNil,
 		DeleteContext: RemoveResourceFromState,
-		Schema:        stringSchemaV1(true),
+		Schema:        passwordSchema,
 		Importer: &schema.ResourceImporter{
 			StateContext: importStringFunc(true),
 		},
+		SchemaVersion: 1,
+		StateUpgraders: []schema.StateUpgrader{
+			{
+				Version: 0,
+				Type:    resourcePasswordV0().CoreConfigSchema().ImpliedType(),
+				Upgrade: resourcePasswordStateUpgradeV0,
+			},
+		},
 	}
+}
+
+func resourcePasswordV0() *schema.Resource {
+	return &schema.Resource{
+		Schema: stringSchemaV1(true),
+	}
+}
+
+func resourcePasswordStateUpgradeV0(ctx context.Context, rawState map[string]interface{}, meta interface{}) (map[string]interface{}, error) {
+	result, ok := rawState["result"].(string)
+	if !ok {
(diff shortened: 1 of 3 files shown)


internal/provider/resource_password.go
```diff
@@ -1,22 +1,93 @@
...
+		Sensitive:   true,
+	}
+
+	create := func(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+		diags := createStringFunc(true)(ctx, d, meta)
+		if diags.HasError() {
+			return diags
+		}
+
+		if d.Get("bcrypt_hash") == "" {
```
bflad, line 27: This attribute will always be empty on creation, so the conditional is unnecessary. 👍

internal/provider/resource_password.go
```diff
@@ -1,22 +1,93 @@
...
 }
+
+func resourcePasswordV0() *schema.Resource {
+	return &schema.Resource{
+		Schema: stringSchemaV1(true),
+	}
+}
+
+func resourcePasswordStateUpgradeV0(ctx context.Context, rawState map[string]interface{}, meta interface{}) (map[string]interface{}, error) {
+	result, ok := rawState["result"].(string)
```
bflad, line 74: To prevent potential panics, I'd recommend a `if rawState == nil` check before here.

internal/provider/resource_password.go
```diff
@@ -1,22 +1,93 @@
...
+	if !ok {
+		return nil, fmt.Errorf("resource password state upgrade failed, result could not be asserted as string: %T", rawState["result"])
+	}
+
+	hash, err := generateHash(result)
+	if err != nil {
+		return nil, fmt.Errorf("resource password state upgrade failed, generate hash error: %v", err)
```
bflad, line 81: Conventionally after Go 1.13, it is recommended to use the `%w` (wrapped error) formatter for errors.

This does leave an interesting question though, what errors can be generated and are they recoverable? If they are not recoverable (e.g. cannot be fixed by retrying), practitioners will be forever stuck with an unusable resource unless they pin to an older provider version or the state upgrade code is "fixed" or updated to skip on the generation error.

internal/provider/resource_password.go
```diff
@@ -1,22 +1,93 @@
...
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"golang.org/x/crypto/bcrypt"
 )
 
 func resourcePassword() *schema.Resource {
+	passwordSchema := stringSchemaV1(true)
+	passwordSchema["bcrypt_hash"] = &schema.Schema{
```
bflad, line 14: Since this new attribute is only handled on Create, e.g. not during Read, it will also need to be handled during Import.

internal/provider/resource_password.go
bendbennett: Added a check.

internal/provider/resource_password.go
bendbennett: Have wrapped the error.

The errors that can be generated relate to:
- [checkCost](https://cs.opensource.google/go/x/crypto/+/46612604:bcrypt/bcrypt.go;l=138) - this should never return an error as we're using _DefaultCost_.
- [io.ReadFull](https://cs.opensource.google/go/x/crypto/+/46612604:bcrypt/bcrypt.go;l=145) - this will return an error if _maxSaltSize_ bytes cannot be read from _rand.Reader_.
- [bcrypt](https://cs.opensource.google/go/x/crypto/+/46612604:bcrypt/bcrypt.go;l=151) - can return error if underlying cipher funcs return error.

internal/provider/resource_password.go
bendbennett: Have added handling of _bcrypt_hash_ to import.

internal/provider/resource_password.go
```diff
@@ -1,22 +1,109 @@
...
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"golang.org/x/crypto/bcrypt"
 )
 
 func resourcePassword() *schema.Resource {
+	create := func(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
```
detro, line 13: Nitpick: Why is this function defined inside inside the other function (lamba)?

I can't spot anything that would require this approach ❓

internal/provider/resource_password.go
```diff
@@ -1,22 +1,109 @@
...
+			{
+				Version: 0,
+				Type:    resourcePasswordV0().CoreConfigSchema().ImpliedType(),
+				Upgrade: resourcePasswordStateUpgradeV0,
+			},
+		},
+	}
+}
+
+func importPasswordFunc() schema.StateContextFunc {
```
detro, line 57: Why do you need a function that returns a function here?

internal/provider/resource_string.go
```diff
@@ -21,9 +24,21 @@ func resourceString() *schema.Resource {
...
-		Schema:        stringSchemaV1(false),
+		Schema:        stringSchemaV1(),
 		Importer: &schema.ResourceImporter{
-			StateContext: importStringFunc(false),
+			StateContext: importStringFunc(),
 		},
 	}
 }
+
+func importStringFunc() schema.StateContextFunc {
```
detro, line 34: Same as above: having this function returning another function seem not necessary.

But I might be missing something.

detro (commented):
Some nitpicks in regards to "function returning functions" that seems unnecessary, but seems like it's pretty much on the finish line.

detro (approved):
After discussing the questions I had in my previous review, I'm providing an approval so you can proceed while you tweak those functions.

internal/provider/resource_password.go
bendbennett: Have removed.

internal/provider/resource_password.go
bendbennett: Have removed.

internal/provider/resource_string.go
bendbennett: Have removed.

detro (approved):
👩‍🍳 😘

PR #255, opened 2022-05-13
Add a hash to random_password with UpgradeState

bendbennett:
This is evaluating whether the changes proposed in https://github.com/hashicorp/terraform-provider-random/pull/73 can be combined with a UpgradeState once the provider has been upgraded to use Framework.

Further test coverage is required.

### DIFF
--- a/internal/provider/diagnostics.go
+++ b/internal/provider/diagnostics.go
@@ -20,3 +20,16 @@ func randomReadError(errMsg string) diag.Diagnostics {
 
 	return diags
 }
+
+func hashGenerationError(errMsg string) diag.Diagnostics {
+	var diags diag.Diagnostics
+
+	diags.AddError(
+		"Hash Generation Error",
+		"While attempting to generate a hash from of the password an error occurred.\n\n"+
+			"Verify that the state contains a populated 'result' field and retry the operation\n\n"+
+			fmt.Sprintf("Original Error: %s", errMsg),
+	)
+
+	return diags
+}
--- a/internal/provider/resource_password.go
+++ b/internal/provider/resource_password.go
@@ -5,6 +5,8 @@ import (
 
 	"github.com/hashicorp/terraform-plugin-framework/diag"
 	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"golang.org/x/crypto/bcrypt"
 )
 
 type resourcePasswordType struct{}
@@ -15,7 +17,17 @@ func (r resourcePasswordType) GetSchema(context.Context) (tfsdk.Schema, diag.Dia
 		"data handling in the [Terraform documentation](https://www.terraform.io/docs/language/state/sensitive-data.html).\n" +
 		"\n" +
 		"This resource *does* use a cryptographic random number generator."
-	return getStringSchemaV1(true, description), nil
+
+	schema := getStringSchemaV1(true, description)
+	schema.Version = 1
+	schema.Attributes["bcrypt_hash"] = tfsdk.Attribute{
+		Description: "A bcrypt hash of the generated random string.",
+		Type:        types.StringType,
+		Computed:    true,
+		Sensitive:   true,
+	}
+
+	return schema, nil
 }
 
 func (r resourcePasswordType) NewResource(_ context.Context, p tfsdk.Provider) (tfsdk.Resource, diag.Diagnostics) {
@@ -29,7 +41,7 @@ type resourcePassword struct {
 }
 
 func (r resourcePassword) Create(ctx context.Context, req tfsdk.CreateResourceRequest, resp *tfsdk.CreateResourceResponse) {
-	createString(ctx, req, resp, true)
+	createPassword(ctx, req, resp)
 }
 
 // Read does not need to perform any operations as the state in ReadResourceResponse is already populated.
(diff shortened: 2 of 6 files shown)


internal/provider/diagnostics.go
```diff
@@ -20,3 +20,16 @@ func randomReadError(errMsg string) diag.Diagnostics {
...
 	return diags
 }
+
+func hashGenerationError(errMsg string) diag.Diagnostics {
+	var diags diag.Diagnostics
+
+	diags.AddError(
+		"Hash Generation Error",
+		"While attempting to generate a hash from of the password an error occurred.\n\n"+
+			"Verify that the state contains a populated 'result' field and retry the operation\n\n"+
```
bflad, line 30: Might be worth mentioning the `terraform state show` command. 👍

internal/provider/string.go
```diff
@@ -283,9 +361,22 @@ func importString(ctx context.Context, req tfsdk.ImportResourceStateRequest, res
...
-	if sensitive {
-		state.ID.Value = "none"
+	diags := resp.State.Set(ctx, &state)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
 	}
+}
+
+func importPassword(ctx context.Context, req tfsdk.ImportResourceStateRequest, resp *tfsdk.ImportResourceStateResponse) {
```
bflad, line 371: Similar to the note on the sdk/v2 version of this, since the new attribute is only handled once on creation, it will also need to be handled during import. It _may_ be desirable to optionally include it in the import ID so folks can have a stable value when doing `terraform state rm` and `terraform import`, although that does introduce yet more complexity. Maybe that can be punted on until there is a specific ask.

detro: Alternatively, could the `bcrypt_hash` value be just computed during import?

After all what matters about it is not strictly having a specific value, but that it's stable once generated.

I'd leave out the option to import it with the ID only if it comes up again, and hopefully deferred to when "import-from-config" is a thing.

internal/provider/resource_password.go
```diff
@@ -15,7 +17,17 @@ func (r resourcePasswordType) GetSchema(context.Context) (tfsdk.Schema, diag.Dia
 		"data handling in the [Terraform documentation](https://www.terraform.io/docs/language/state/sensitive-data.html).\n" +
 		"\n" +
 		"This resource *does* use a cryptographic random number generator."
-	return getStringSchemaV1(true, description), nil
+
+	schema := getStringSchemaV1(true, description)
```
detro, line 21: Nit: the name of the function `getStringSchemaV1` makes one thing that the schema that comes out of it is `V1`.

But it's not right? It's `V0`. `V1` is what you are creating now, by adding `bcrypt_hash`.

IMHO: I'd try to preserve consistency and have 2 functions able to provide the schema, and the version correctly respected in their name.

internal/provider/diagnostics.go
bendbennett: Added comment to use `terraform state show`.

internal/provider/string.go
bendbennett: I've updated `importPassword` to handle `bcrypt_hash` during import. 

I'm inclined to defer handling of a passed `bcrypt_hash` during import until/unless it is specifically requested.

internal/provider/resource_password.go
bendbennett: Good point. Have refactored.

detro (approved):

internal/provider/resource_password.go
```diff
@@ -47,9 +47,256 @@ func (r resourcePassword) Delete(ctx context.Context, req tfsdk.DeleteResourceRe
...
 
 func (r resourcePassword) ValidateConfig(ctx context.Context, req tfsdk.ValidateResourceConfigRequest, resp *tfsdk.ValidateResourceConfigResponse) {
 	validateLength(ctx, req, resp)
 }
+
+func (r resourcePassword) UpgradeState(context.Context) map[int64]tfsdk.ResourceStateUpgrader {
+	return map[int64]tfsdk.ResourceStateUpgrader{
+		0: {
+			StateUpgrader: migratePasswordStateV0toV1,
+		},
```
bflad, line 61: Any particular reason to opt for handling `RawState` in the `StateUpgrader` versus specifying the version 0 schema here in the `PriorSchema` field? If you use `PriorSchema`, you can call `req.PriorState.Get()` to populate the already defined data type.

bendbennett: Yep, I went the route of using `RawState` as `req.State` is nil in the `tfsdk.UpgradeResourceStateRequest` and I can't see a `PriorState` field in `req`.

bflad: Sorry I meant, [`UpgradeResourceStateRequest.State`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/tfsdk#UpgradeResourceStateRequest). It get's filled in when [`ResourceStateUpgrader.PriorSchema`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/tfsdk#ResourceStateUpgrader) is defined. It should make the logic a lot easier.

bendbennett: Thanks for the clarification. I've refactored to use `StateUpgrader` with `PriorSchema`.

bflad (approved):
Actually clicking the button 😅

internal/provider/resource_pasword_test.go
```diff
@@ -86,29 +107,66 @@ func TestAccResourcePasswordMin(t *testing.T) {
...
+		OverrideSpecial: types.String{Value: "!#$%\u0026*()-_=+[]{}\u003c\u003e:?"},
+		Result:          types.String{Value: "DZy_3*tnonj%Q%Yx"},
+	}
+
+	actual := PasswordModelV1{}
+	resp.State.Get(context.Background(), &actual)
+
+	err := bcrypt.CompareHashAndPassword([]byte(actual.BcryptHash.Value), []byte(actual.Result.Value))
+	if err != nil {
+		t.Error(err)
```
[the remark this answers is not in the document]
bendbennett: Missed this before merging but have added into https://github.com/hashicorp/terraform-provider-random/pull/249

PR #258, opened 2022-05-24
Deprecating number in favour of numeric

bendbennett:
In order to align attribute naming, `number` is being deprecated and `numeric` has been added.
This gives the following attribute pairs:
```
numeric (bool)
min_numeric (int)
lower (bool)
min_lower (int)
upper (bool)
min_upper (int)
..............
```

In order to allow both `number` and `numeric` to co-exist and to be able to keep them in-sync and have them default to `true` when they are absent from config, the attribute schema for number was updated to remove `Default` and add `Computed`. 

The `Default` behaviour was then replicated by using `CustomizeDiffFunc(s)` on `resourcePassword` and `resourceString` and the values of `number` and `numeric` were kept in-sync also by using `CustomizeDiffFunc(s)`.


internal/provider/string.go
```diff
@@ -257,3 +322,67 @@ func generateRandomBytes(charSet *string, length int) ([]byte, error) {
...
+
+func resourceStateUpgradeAddNumeric(_ context.Context, rawState map[string]interface{}, _ interface{}, resourceName string) func(_ context.Context, rawState map[string]interface{}, _ interface{}) (map[string]interface{}, error) {
+	return func(_ context.Context, rawState map[string]interface{}, _ interface{}) (map[string]interface{}, error) {
+		if rawState == nil {
+			return nil, fmt.Errorf("resource %s state upgrade failed, state is nil", resourceName)
+		}
+
+		number, ok := rawState["number"].(bool)
+		if !ok {
+			return nil, fmt.Errorf("resource %s state upgrade failed, number could not be asserted as bool: %T", resourceName, rawState["number"])
```
detro, line 334:
```suggestion
			return nil, fmt.Errorf("resource %s state upgrade failed, 'number' is not a boolean: %T", resourceName, rawState["number"])
```

Just thinking it could be easier to understand for the practitioner. Feel free to discard.

internal/provider/string.go
```diff
@@ -257,3 +322,67 @@ func generateRandomBytes(charSet *string, length int) ([]byte, error) {
...
+		}
+
+		rawState["numeric"] = number
+
+		return rawState, nil
+	}
+}
+
+func customDiffIfValue(key string) func(context.Context, *schema.ResourceDiff, interface{}) error {
```
detro, line 343: I think this function needs a bit of godoc to explain what's needed for. I remember a bit of what you explained verbally, but I will forget in a few hours.

internal/provider/string.go
```diff
@@ -257,3 +322,67 @@ func generateRandomBytes(charSet *string, length int) ([]byte, error) {
...
+				if err != nil {
+					return err
+				}
+			}
+			return nil
+		},
+	)
+}
+
+func customDiffIfValueChangeNumber() func(context.Context, *schema.ResourceDiff, interface{}) error {
```
detro, line 366: Same as per function above (and below).

Just something to explain the whole "if `number` has changed, we want to prograpage to `numeric`" and vice-versa.

internal/provider/string.go
```diff
@@ -257,3 +322,67 @@ func generateRandomBytes(charSet *string, length int) ([]byte, error) {
 func readNil(_ context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
 	return nil
 }
+
+func resourceStateUpgradeAddNumeric(_ context.Context, rawState map[string]interface{}, _ interface{}, resourceName string) func(_ context.Context, rawState map[string]interface{}, _ interface{}) (map[string]interface{}, error) {
```
detro, line 326: I think the linter is complaining about the `rawState` in the signature.

detro (commented):
Looking great. I just left a few comments about... adding godoc comments :)

The problem you solved is pretty complex, so it's actually crucial that your use of `CustomizeDiff` is accompanied by godoc that will help the next figure out what problem you solved there.

internal/provider/string.go
bendbennett: Agreed. Fixed.

internal/provider/string.go
bendbennett: Agreed. Added.

internal/provider/string.go
bendbennett: Agreed. Added.

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 3.3.0 (Unreleased)
+
+ENHANCEMENTS:
+
+* resource/random_password: `number` is now deprecated and `numeric` has been added to align attribute naming. `number` will be removed in the next major release ([#258](https://github.com/hashicorp/terraform-provider-random/pull/258)).
+* resource/random_string: `number` is now deprecated and `numeric` has been added to align attribute naming. `number` will be removed in the next major release ([#258](https://github.com/hashicorp/terraform-provider-random/pull/258)).
+
```
detro, line 7: 👩‍🍳 💋

internal/provider/resource_password_test.go
```diff
@@ -0,0 +1,514 @@
...
+					resource.TestCheckResourceAttr("random_password.default", "number", "true"),
+					resource.TestCheckResourceAttr("random_password.default", "numeric", "true"),
+				),
+			},
+		},
+	})
+}
+
+// TestAccResourcePassword_StateUpgraders covers the state upgrades from v0 to V2 and V1 to V2.
+// This includes the addition of bcrypt_hash and numeric attributes.
```
detro, line 166: ❤️

detro (approved):
LGTM

internal/provider/string.go
bflad: The naming and single parameter confused me a little bit here at first glance! I wonder if it is worth renaming and making this slightly more generic (not sure how useful it'd be outside this provider, but I think the naming and signature will make it more clear what the intent is), e.g.

```go
func planDefaultIfAllNull(default interface{}, keys ...string) []schema.CustomizeDiffFunc {
  var result []schema.CustomizeDiffFunc

  for _, key := range keys {
    result = append(result, customdiff.IfValue(/*...*/))
  }

  return result
}
```

With the callers then being `planDefaultIfAllNull(true, "number", "numeric")...`

bflad (commented):
All the new testing 😍

internal/provider/resource_string_test.go
```diff
@@ -74,6 +77,336 @@ func TestAccResourceStringMin(t *testing.T) {
...
+		afterStateUpgrade   []resource.TestCheckFunc
+	}{
+		{
+			name: "%s number is absent",
+			configBeforeUpgrade: `resource "random_string" "default" {
+						length = 12
+					}`,
+			beforeStateUpgrade: []resource.TestCheckFunc{
+				resource.TestCheckResourceAttrSet("random_string.default", "number"),
+				resource.TestCheckResourceAttr("random_string.default", "number", "true"),
```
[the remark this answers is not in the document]
bendbennett: Have removed redundant checks for `TestCheckResourceAttrSet`.

internal/provider/string.go
```diff
@@ -257,3 +323,78 @@ func generateRandomBytes(charSet *string, length int) ([]byte, error) {
 func readNil(_ context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
 	return nil
 }
+
+func resourceStateUpgradeAddNumeric(resourceName string) func(_ context.Context, rawState map[string]interface{}, _ interface{}) (map[string]interface{}, error) {
+	return func(_ context.Context, rawState map[string]interface{}, _ interface{}) (map[string]interface{}, error) {
+		if rawState == nil {
+			return nil, fmt.Errorf("resource %s state upgrade failed, state is nil", resourceName)
```
bflad, line 330: Does Terraform CLI not already make it clear which resource address failed, which would inherently include the resource name? If not, maybe we should raise a feature request upstream so adding the resource name, which may not be the most helpful in large configurations, is not necessary in provider code.

bflad (approved):
Open question above ☝️ but approving to unblock you.

internal/provider/string.go
bendbennett: I've just checked this. Terraform CLI _does_ make it clear which resource address failed. For instance,
```bash
│ Error: state upgrade failed, state is nil
│ 
│   with random_password.password,
│   on resource.tf line 1, in resource "random_password" "password":
│    1: resource "random_password" "password" {
```

Consequently, I have removed usage of the resource name from the error messages in and refactored to replace `resourceStateUpgradeAddNumeric` func with `resourcePasswordStringStateUpgradeV1`.

bflad (approved):
Looks good to me 🚀

Issue #63, opened 2019-06-14
General random bytes, perhaps base64 encoded?


deitch:
Are there any plans to offer a resource to generate generic random bytes, perhaps base64-encoded if storage/representation is an issue?

* `random_string` generates the requested number of bytes, but is limited to the alphanumeric and defined special character range. By default, that is ~83 potential bytes (26 lcase, 26 ucase, 10 digits, and another 21 special characters)
* `random_integer` does provide a theoretical full range, but if you want to generate anything useful, e.g. 16 or 32 bytes of randomness, I don't know if it will get there.
* `random_uuid` has 16 bytes and is randomly generated, but the format returned is the normal UUID string, which would require some hoops to jump through (if even possible using tf interpolation) to get the raw bytes.

I could see a useful case, something like:

```
resource "random_bytes" b {
  length = 32
}

output "randbytes" {
  value = random_bytes.b.bytes // gives actual bytes, which may be unprintable, but are passable
}
output "randb64" {
  value = random_bytes.b.base64 // gives a string
}
```

Trying to find a way to get the same result in the meantime (I haven't figured it out yet), but this would be quite useful.

deitch:
I realize that using `random_integer` that way is not really useful. An alternative might be to use `random_integer` 16 times, but that still gives a difficult to manage result, and appears to use `math/rand` rather than `crypt/rand` (unlike uuid and string):

```terraform
provider "random"{}

resource "random_integer" test {
  count = 16
  min = 1
  max = 256
}

output "int" {
  value = "${random_integer.test.*.result}"
}
```

The output is not easy to work with in HCL, either. Plenty easy in go though. :-)

deitch:
I managed to do this. I don't love it, would be nice if it were a basic resource.

```terraform
provider "random"{}

resource "random_string" test {
  length = 16
[...]

PR #272, opened 2022-07-08
Add new resource `random_bytes` to generate an array of random bytes

Socolin:
Intended to be used as key or secret

Fix https://github.com/hashicorp/terraform-provider-random/issues/63


Socolin:
@bendbennett Hello, is there any chance to see this PR merge (or if it's not something you want, rejected) soon ? I need this I would like to know if at least it's something you are considering or if I should instead create my own provider ?

bendbennett:
@Socolin apologies for delay in responding on this. I'm afraid that I don't have an ETA at this time but I will endeavour to get some feedback to you.

dnlopes:
I'm also waiting badly for this PR to get merged. Have a comment in my code base pointing here 😄

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## Upcoming
+
+NEW FEATURES:
+
+* `random_bytes` generates an array of random bytes that is intended to be used as key or secret ([#63](https://github.com/hashicorp/terraform-provider-random/issues/63))
+
```
bendbennett, line 6:
```suggestion
## 3.5.0 (unreleased)

NEW FEATURES:

* resource/random_bytes: Generates an array of random bytes that is intended to be used as key or secret ([#272](https://github.com/hashicorp/terraform-provider-random/pull/272))

```

bendbennett (commented):
Thanks for the contribution.

Could you rebase on `main`?
Looks like this PR is using `v0.9.0` of `terraform-plugin-framework`.

sebhaub:
nice - this PR is exactly what i currently miss in the random provider. 
@bendbennett is there something missing / preventing to merge this ?

jmgilman:
I'd be willing to help out if it means getting this merged sooner. A huge need for us, thanks!

Socolin:
@bendbennett Do you have an ETA on this ?

bookshelfdave:
We hope to release this in the next few weeks, it's queued behind some other work that we have planned.

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 3.5.0 (unreleased)
+
+NEW FEATURES:
+
+* `random_bytes` generates an array of random bytes that is intended to be used as key or secret ([#63](https://github.com/hashicorp/terraform-provider-random/issues/63))
+
```
bendbennett, line 6:
```suggestion
```

We have recently switched to using [Changie](https://changie.dev/) for changelog and version management.

Could you remove this entry from CHANGELOG.md and once Changie is [installed](https://changie.dev/guide/installation/) run the following:

```shell
changie new
? Kind: Features
Body: resource/random_bytes: New resource that generates an array of random bytes intended to be used as key or secret ([#63](https://github.com/hashicorp/terraform-provider-random/issues/63)).
Issue/PR Number: 272
```

internal/provider/resource_bytes.go
```diff
@@ -0,0 +1,169 @@
...
+	ID           types.String `tfsdk:"id"`
+	Length       types.Int64  `tfsdk:"length"`
+	Keepers      types.Map    `tfsdk:"keepers"`
+	ResultBase64 types.String `tfsdk:"result_base64"`
+	ResultHex    types.String `tfsdk:"result_hex"`
+}
+
+func bytesSchemaV3() tfsdk.Schema {
+	return tfsdk.Schema{
+		Version: 3,
```
bendbennett, line 124: As this is the first version of the schema for the `bytes` resource, we could use `Version: 0`.

internal/provider/resource_bytes.go
```diff
@@ -0,0 +1,169 @@
...
+
+type bytesModelV0 struct {
+	ID           types.String `tfsdk:"id"`
+	Length       types.Int64  `tfsdk:"length"`
+	Keepers      types.Map    `tfsdk:"keepers"`
+	ResultBase64 types.String `tfsdk:"result_base64"`
+	ResultHex    types.String `tfsdk:"result_hex"`
+}
+
+func bytesSchemaV3() tfsdk.Schema {
```
bendbennett, line 122: If `Version: 0` is used for the schema the function name could be `bytesSchemaV0()`.

bendbennett (commented):
@Socolin thank you for your contribution.

Would you be happy to add a couple more tests to verify:

- Behaviour of validator on `length` attribute is as expected.
- Altering `length` forces a replacement.

docs/resources/bytes.md
```diff
@@ -0,0 +1,51 @@
+---
+page_title: "random_bytes Resource - terraform-provider-random"
+subcategory: ""
+description: |-
+  The resource random_bytes generates an array of bytes that is intended to be used as key or secret.
+---
+
+# random_password (Resource)
```
maaarghk, line 8: should be `random_bytes`

docs/resources/bytes.md
```diff
@@ -0,0 +1,51 @@
...
+- `result_base64` (String, Sensitive) The generated bytes presented in base64 string format.
+- `result_hex` (String, Sensitive) The generated bytes presented in hex string format.
+- `id` (String) A static value used internally by Terraform, this should not be referenced in configurations.
+
+## Import
+
+Import is supported using the following syntax:
+
+```shell
+terraform import random_bytes.secret 8/fu3q+2DcgSJ19i0jZ5Cw==
```
maaarghk, line 50: mentioning this is base64 is probably justified here

internal/provider/resource_bytes.go
```diff
@@ -0,0 +1,169 @@
...
+				Sensitive:   true,
+			},
+			"result_hex": {
+				Description: "The generated bytes presented in hex string format.",
+				Type:        types.StringType,
+				Computed:    true,
+				Sensitive:   true,
+			},
+			"id": {
+				Description: "The generated bytes presented in string format.",
```
maaarghk, line 163: at a cursory glance I don't think this is accurate, "id" appears to always be set to "none". Other providers have this described as "A static value used internally by Terraform, this should not be referenced in configurations."

Should `result` be added to actually return the generated bytes in string format? probably a question for @bendbennett - not sure how the state file is expected to behave with binary data in it - or whether it could be achieved in the `Read` function - or if anyone would ever use it anyway

internal/provider/resource_bytes.go
```diff
@@ -0,0 +1,169 @@
...
+				fmt.Sprintf("Original Error: %s", err),
+		)
+		return
+	}
+
+	var state bytesModelV0
+
+	state.ID = types.StringValue("none")
+	state.Length = types.Int64Value(int64(len(bytes)))
+	state.ResultBase64 = types.StringValue(base64.StdEncoding.EncodeToString(bytes))
```
maaarghk, line 103: nit: I don't think re-encoding is required here, can just do `req.ID`

internal/provider/resource_bytes_test.go
```diff
@@ -0,0 +1,48 @@
...
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestMatchResourceAttr("random_bytes.basic", "result_base64", regexp.MustCompile(`^[A-Za-z/+\d]{43}=$`)),
+					resource.TestMatchResourceAttr("random_bytes.basic", "result_hex", regexp.MustCompile(`^[a-f\d]{64}$`)),
+					resource.TestCheckResourceAttr("random_bytes.basic", "length", "32"),
+				),
+			},
+			{
+				// Usage of ImportStateIdFunc is required as the value passed to the `terraform import` command needs
+				// to be the password itself, as the password resource sets ID to "none" and "result" to the password
+				// supplied during import.
```
maaarghk, line 29: nit: the above comments should refer to bytes rather than password I suppose

internal/provider/resource_bytes_test.go
```diff
@@ -0,0 +1,48 @@
...
+					id := "random_password.basic"
+					rs, ok := s.RootModule().Resources[id]
+					if !ok {
+						return "", fmt.Errorf("not found: %s", id)
+					}
+					if rs.Primary.ID == "" {
+						return "", fmt.Errorf("no ID is set")
+					}
+
+					return rs.Primary.Attributes["result"], nil
```
maaarghk, line 40: I guess this should be updated to have `random_bytes.basic` for the id and to return `result_base64`?

maaarghk (changes requested):

internal/provider/resource_bytes.go
Socolin: I did not set `result` to avoid confusion, since the result is always encoded. But I could add it, we just need to choose if the default is base64 or hex (or other ?)

Socolin:
Thanks for the comments, I pushed a fix for them.

maaarghk:
> Thanks for the comments, I pushed a fix for them.

Think everything I left looks good, although maybe I obscured a request from Ben for two more tests to be added :)

Socolin:
> > Thanks for the comments, I pushed a fix for them.
> 
> Think everything I left looks good, although maybe I obscured a request from Ben for two more tests to be added :)

Thanks I forgot about this, I'm checking this

Socolin:
I added 2 tests. I noticed that there are many tests for `keepers` in the `random_password` resource. Should there also be tests for random_bytes? If so, could you share tests for both resources instead of copy/pasting them?

.changes/unreleased/FEATURES-20230213-213454.yaml
```diff
@@ -0,0 +1,6 @@
+kind: FEATURES
+body: 'resource/random_bytes: New resource that generates an array of random bytes
+  intended to be used as key or secret ([#63](https://github.com/hashicorp/terraform-provider-random/issues/63)).'
```
bendbennett, line 3:
```suggestion
  intended to be used as key or secret'
```

Apologies, believe I suggested including a link in the changie file. This is not needed as changie will generate a link on the basis of the specified `Issues`.

internal/provider/resource_bytes_test.go
```diff
@@ -0,0 +1,94 @@
...
+			},
+			{
+				ProtoV5ProviderFactories: protoV5ProviderFactories(),
+				Config: `resource "random_bytes" "test" {
+					length = 2
+				}`,
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("random_bytes.test", "length", "2"),
+					testExtractResourceAttr("random_bytes.test", "result_base64", &bytes2),
+					testExtractResourceAttr("random_bytes.test", "result_base64", &bytes2),
```
bendbennett, line 87:
```suggestion
```

Looks like the same value is being extracted twice.

internal/provider/resource_bytes.go
bendbennett: As the base64 and hex encodings of `result` are available it doesn't seem like there is a need for `result` as @Socolin indicates it would need to be encoded.

bendbennett:
> I added 2 tests. I noticed that there are many tests for `keepers` in the `random_password` resource. Should there also be tests for random_bytes? If so, could we share tests for both resources instead of copy/pasting them?

Believe some tests around `keepers` would be useful. I think code duplication for tests is ok in this case.

internal/provider/resource_bytes_test.go
Socolin: Line was duplciated by error, I fix

Socolin:
I added tests for `keepers`  but 2 test don't pass and I don't understand why

`TestAccResourceBytes_Keepers_Keep_NullMapToNullValue`
`TestAccResourceBytes_Keepers_Keep_EmptyMapToNullValue`

Do you know why ? @bendbennett

internal/provider/resource_bytes.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+// Read does not need to perform any operations as the state in ReadResourceResponse is already populated.
+func (r *bytesResource) Read(context.Context, resource.ReadRequest, *resource.ReadResponse) {
+}
+
+// Update is intentionally left blank as all required and optional attributes force replacement of the resource
+// through the RequiresReplace AttributePlanModifier.
+func (r *bytesResource) Update(context.Context, resource.UpdateRequest, *resource.UpdateResponse) {
+}
```
bendbennett, line 83:
```suggestion
func (r *bytesResource) Update(context.Context, resource.UpdateRequest, *resource.UpdateResponse) {
	var model bytesModelV0

	resp.Diagnostics.Append(req.Plan.Get(ctx, &model)...)

	if resp.Diagnostics.HasError() {
		return
	}

	resp.Diagnostics.Append(resp.State.Set(ctx, &model)...)
}
```

internal/provider/resource_bytes.go
```diff
@@ -0,0 +1,166 @@
...
+					int64planmodifier.RequiresReplace(),
+				},
+				Validators: []validator.Int64{
+					int64validator.AtLeast(1),
+				},
+			},
+			"result_base64": schema.StringAttribute{
+				Description: "The generated bytes presented in base64 string format.",
+				Computed:    true,
+				Sensitive:   true,
```
bendbennett, line 153:
```suggestion
				Sensitive:   true,
				PlanModifiers: []planmodifier.String{
					stringplanmodifier.UseStateForUnknown(),
				},
```

internal/provider/resource_bytes.go
```diff
@@ -0,0 +1,166 @@
...
+			},
+			"result_hex": schema.StringAttribute{
+				Description: "The generated bytes presented in hex string format.",
+				Computed:    true,
+				Sensitive:   true,
```
bendbennett, line 158:
```suggestion
				Sensitive:   true,
				PlanModifiers: []planmodifier.String{
					stringplanmodifier.UseStateForUnknown(),
				},				
```

internal/provider/resource_bytes.go
```diff
@@ -0,0 +1,166 @@
...
+			},
+			"id": schema.StringAttribute{
+				Description: "A static value used internally by Terraform, this should not be referenced in configurations.",
+				Computed:    true,
```
bendbennett, line 162:
```suggestion
				Computed:    true,
				PlanModifiers: []planmodifier.String{
					stringplanmodifier.UseStateForUnknown(),
				},				
```

bendbennett:
> I added tests for `keepers` but 2 test don't pass and I don't understand why
> 
> `TestAccResourceBytes_Keepers_Keep_NullMapToNullValue` `TestAccResourceBytes_Keepers_Keep_EmptyMapToNullValue`

The issue here is that there is a discrepancy between the `plan` and the `state` during the `Update`.

The `mapplanmodifiers.RequiresReplaceIfValuesNotNull()` does not mark `keepers` as `RequiresReplace` because `keepers` is updated to:
```
keepers = {
	"key" = null
}
```  

This results in a call to the `Update` function on the resource. As this function is empty, the `state` that is passed in via `resource.UpdateRequest` is unaltered in `resource.UpdateResponse`. Terraform detects that there is a mismatch between the `plan` and the `state` and indicates this through:
```
        Error: Provider produced inconsistent result after apply
        
        When applying changes to random_bytes.test, provider
        "provider[\"registry.terraform.io/hashicorp/random\"]" produced an unexpected
        new value: .keepers: was
        cty.MapVal(map[string]cty.Value{"key":cty.NullVal(cty.String)}), but now
        null.
```

I've added some comments into the PR with the required code changes.

Socolin:
Ok thanks for the explanation, it should now be good to merge I think.

dhermes:
> Ok thanks for the explanation, it should now be good to merge I think.

This is great to see! Cheers everyone

Socolin:
I just saw the build failed (https://github.com/hashicorp/terraform-provider-random/actions/runs/4179417990), I think the reason is that I did not ran `make generate`, it's now done.

internal/provider/resource_bytes.go
```diff
@@ -0,0 +1,183 @@
...
+			"used as secret or keys.",
+		Attributes: map[string]schema.Attribute{
+			"keepers": schema.MapAttribute{
+				Description: "Arbitrary map of values that, when changed, will trigger recreation of " +
+					"resource. See [the main provider documentation](../index.html) for more information.",
+				ElementType: types.StringType,
+				Optional:    true,
+				PlanModifiers: []planmodifier.Map{
+					mapplanmodifiers.RequiresReplaceIfValuesNotNull(),
+				},
```
bflad, line 146: Drive-by comment: This new resource could use the regular `mapplanmodifier.RequiresReplace()` as there shouldn't be any prior state storage from a terraform-plugin-sdk based resource which would've collapsed a map of null values into a null map (the intended usage of the custom plan modifier was to prevent replacing existing resources stored with the prior SDK's invalid state). It may be worth an additional code comment here on the design decision either which way, whether to keep consistency with the other resources or not (so all `keepers` changes including adding `null` values replace the resource).

Socolin: @bendbennett should I change this or keep it that way ?

bendbennett: @Socolin just for the sake of consistency with the other resources, I think I'd be inclined to use `mapplanmodifier.RequiresReplace()` and add a code comment as @bflad suggests.

Socolin: What the comment should say ?  something like ?:

`mapplanmodifier.RequiresReplace()` could be used, but for consistency with other resources we use  `mapplanmodifiers.RequiresReplaceIfValuesNotNull() ` 

or do you have any other suggestion ? @bendbennett

bendbennett: @Socolin it might be useful to refer to the fact that _random_bytes_ was not available as a a resource on the _random_ provider until after the migration from SDKv2 to the Framework. 

Perhaps something like:
> mapplanmodifiers.RequiresReplaceIfValuesNotNull() has been used for consistency with other resources but mapplanmodifier.RequiresReplace() could have been used as there shouldn't be any prior state storage from a terraform-plugin-sdk based resource which would've collapsed a map of null values into a null map.

Socolin: Ok, thanks.

Socolin:
I fixed the last review comment

internal/provider/resource_bytes.go
```diff
@@ -0,0 +1,187 @@
...
+	if resp.Diagnostics.HasError() {
+		return
+	}
+}
+
+// Read does not need to perform any operations as the state in ReadResourceResponse is already populated.
+func (r *bytesResource) Read(context.Context, resource.ReadRequest, *resource.ReadResponse) {
+}
+
+// Update is intentionally left blank as all required and optional attributes force replacement of the resource
```
bendbennett, line 81: Believe this comment can be removed after the last set of changes.

internal/provider/resource_bytes.go
```diff
@@ -0,0 +1,187 @@
...
+		return
+	}
+
+	var state bytesModelV0
+
+	state.ID = types.StringValue("none")
+	state.Length = types.Int64Value(int64(len(bytes)))
+	state.ResultBase64 = types.StringValue(req.ID)
+	state.ResultHex = types.StringValue(hex.EncodeToString(bytes))
+	state.Keepers = types.MapValueMust(types.StringType, nil)
```
bendbennett, line 116: I think perhaps we want to use `state.Keepers = types.MapNull(types.StringType)` here otherwise there will be a forced update following import if the configuration does not contain an empty `keepers` attribute. For example,

```terraform
resource "random_bytes" "basic" {
  length = 16
  keepers = {}
}
```

Could you add a test to ensure that following import when apply is run there are no planned changes?

Socolin: I added a test, `types.MapNull(types.StringType)` does not seems to changes the outcome, but with the test I think it's validating this scenario

dominicroessner:
Any update on when this will be released? We’re looking to use this also.

internal/provider/resource_bytes_test.go
```diff
@@ -0,0 +1,643 @@
...
+				),
+			},
+			{
+				ResourceName: "random_bytes.basic",
+				Config: `resource "random_bytes" "basic" {
+							length = 32
+						}`,
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("random_bytes.basic", "result_base64", "hkvbcU5f8qGysTFhkI4gzf3yRWC1jXW3aRLCNQFOtNw="),
+				),
```
bendbennett, line 74: Using `types.MapNull(types.StringType)` rather than `types.MapValueMust(types.StringType, nil)` prevents an in-place update if the `random_bytes` resource is imported and then `terraform apply` is run.

Could you amend the test as follows:

```suggestion
				PlanOnly: true,
```

If you try running this test whilst using `types.MapValueMust(types.StringType, nil)` you can see that it fails with the following error:

```          # random_bytes.basic will be updated in-place
          ~ resource "random_bytes" "basic" {
                id            = "none"
              - keepers       = {} -> null
                # (3 unchanged attributes hidden)
            }
        
        Plan: 0 to add, 1 to change, 0 to destroy.
```

internal/provider/resource_bytes_test.go
```diff
@@ -0,0 +1,643 @@
...
+							length = 32
+						}`,
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("random_bytes.basic", "result_base64", "hkvbcU5f8qGysTFhkI4gzf3yRWC1jXW3aRLCNQFOtNw="),
+					resource.TestCheckResourceAttr("random_bytes.basic", "result_hex", "864bdb714e5ff2a1b2b13161908e20cdfdf24560b58d75b76912c235014eb4dc"),
+					resource.TestCheckResourceAttr("random_bytes.basic", "length", "32"),
+				),
+			},
+			{
+				ResourceName: "random_bytes.basic",
```
bendbennett, line 68:
```suggestion
```

bendbennett: I think you can delete `ResourceName: "random_bytes.basic",`, I don't think it's required.

Socolin: > I think you can delete `ResourceName: "random_bytes.basic",`, I don't think it's required.

I forgot to push my commit, it's fixed

Socolin:
I rebased the branch on `main` too

bendbennett (approved):

ohemelaar:
Hello,
I've been lurking at this PR for a while now, I'm quite interested in using this. I see the PR is approved, is there a specific release schedule that's blocking the merge now?

dominicroessner:
Any update or timeline on when this might go in would really be appreciated. Waiting for this as others. Thank you.

Socolin:
Hello,

@bendbennett What is the next steps for this to be merged and released ?

bendbennett:
Hi @Socolin 👋

I'd like to get an additional approval of the PR prior to merging as these changes are a significant alteration to the _random_ utility provider. The efforts of the team are currently focussed elsewhere at the moment but we will endeavour to focus some attention on this as soon as we're able.

GerardSoleCa:
Hi @bendbennett, another interested party here. I'd really appreciate to have this in the random provider. We are currently generating tokens for akamai CDN which require hex strings, and the only option I saw is to use the random_id resource, but that's a really insecure option since output is not sensitive :(

Hope you get soon a second approval for this PR.

Cheers!

Socolin:
@bendbennett Hello, do you have any ETA or "Not before" date for this ?
I would like to know to take some decision about this, I might publish this in another provider so I can start using it, but I will not do this if it's close to be merged.

bendbennett:
Hi @Socolin and @GerardSoleCa 👋 

Apologies for the delay. The focus of the team is currently elsewhere at the moment so it's hard to give an ETA.

Socolin:
To anyone looking for this, I published this in a separate provider

https://registry.terraform.io/providers/Socolin/randombyte/latest

bendbennett:
Hi @Socolin 👋 

Apologies for the delay in getting back to you. We have recently discussed your PR, and, if you're still amenable, we would like to move forward with the addition of the `random_bytes` resource.

The following suggestions came out of our discussion:

- Could `result_base64`, and `result_hex` be renamed `base64`, and `hex`, respectively?
- Could `id` be removed? Just to clarify, it is not used by practitioners, it is not required for the provider supported versions of Terraform (>= 0.12), and it is no longer needed for testing purposes (i.e., when the [terraform-plugin-testing](https://github.com/hashicorp/terraform-plugin-testing) module is used).
- With the addition of `random_bytes`, it would likely be valuable to update the documentation for both `random_id`, and `random_bytes` to provide some clarification around the use-case for using one or other of these two resources. The most obvious differentiator being when the output would be considered sensitive, and should not be displayed in CLI console output.

Thank you again for your contribution, and sorry it has taken a while for us to get back to you.

bendbennett:
Hi @Socolin 👋 

I've incoporated all of your commits into [Add random_bytes resource](https://github.com/hashicorp/terraform-provider-random/pull/494#top).

If you'd rather update this PR with the suggestions in this [comment](https://github.com/hashicorp/terraform-provider-random/pull/272#issuecomment-1824621139), then that would be great. If not, then we'll move forward with [Add random_bytes resource](https://github.com/hashicorp/terraform-provider-random/pull/494#top).

Many thanks for all your work on this, and apologies for the delay.

Socolin:
Hello,

I'm working on other project right now, you can move forward with your PR, Thanks :)

PR #285, opened 2022-07-26
Fix typo

andygarfield:


yardenshoham (approved):

bendbennett:
Hi @andygarfield :wave: Thank you for the fix for the docs.

The contents of the `uuid.md` file that have been amended are drawn from the `Description` field in the Schema.

Would you be happy to update the [Description](https://github.com/hashicorp/terraform-provider-random/blob/main/internal/provider/resource_uuid.go#L28) field and then run `make generate` to recreate the docs? Thanks.

bflad (approved):
Looks good to me, thank you @andygarfield 🚀

bflad:
Closing only because this change request was not updated as noted above for a very long time period and subsequently done in https://github.com/hashicorp/terraform-provider-random/pull/426. Thank you for the suggestion though, @andygarfield 👍

Issue #296, opened 2022-08-22
Remove unneeded code from attribute plan modifier


bendbennett:
### Terraform CLI and Provider Versions

Terraform v1.2.5

### Use Cases or Problem Statement

During migration from SDKv2 to the Framework, an [attribute plan modifier](https://github.com/hashicorp/terraform-provider-random/blob/86d23615282436a5c5e94518db752dc6b8e76fa7/internal/planmodifiers/attribute.go#L105) that mimics the behaviour of a [function used in CustomizeDiff in v3.3.2](https://github.com/hashicorp/terraform-provider-random/blob/v3.3.2/internal/provider/string.go#L385) was created. The attribute plan modifier contains some unneeded code.

### Proposal

Remove unneeded code from the [attribute plan modifier](https://github.com/hashicorp/terraform-provider-random/blob/86d23615282436a5c5e94518db752dc6b8e76fa7/internal/planmodifiers/attribute.go#L105).

### How much impact is this issue causing?

Low

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #297, opened 2022-08-22
Remove unneeded code from plan modifier

bendbennett:
Closes: #296


bflad (approved):
Looks good to me as those two variables were based off `req.Config` anyways 🚀 

Doesn't need to be part of this change, but heads up that `req.Config.GetAttribute(ctx, path.Root("numeric"), &numericConfig)` does not have its diagnostics saved (missing `diags = ...`) which might be good to fix here and maybe something we should consider the framework design or add it to a potential static analysis checking mechanism in the future.

Issue #302, opened 2022-08-31
Upgrade v2.2.1 -> v3.4.0 forces replacement


blgm:
### Terraform CLI and Provider Versions

Terraform v1.2.8
on darwin_amd64
+ provider registry.terraform.io/hashicorp/random v3.4.0

### Terraform Configuration

```terraform
terraform {
  required_providers {
    random = {
      source = "hashicorp/random"
#      version = "3.3.2"
      version = "3.4.0"
#      version = "2.2.1"
    }
  }
}

resource "random_password" "password" {
  length           = 31
  override_special = "~_-."
  min_upper        = 2
  min_lower        = 2
  min_special      = 2
}
```

### Expected Behavior

When a password was created with v2.2.1, I expect to be able to upgrade to v3.4.0 and for there to be no changes to the password.

### Actual Behavior

When a password was created with v2.2.1, when I upgrade to v3.4.0 the password is replaced:
```
random_password.password: Refreshing state... [id=none]

Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
-/+ destroy and then create replacement

Terraform will perform the following actions:

  # random_password.password must be replaced
-/+ resource "random_password" "password" {
      ~ bcrypt_hash      = (sensitive value)
      ~ id               = "none" -> (known after apply)
      ~ min_upper        = 0 -> 2 # forces replacement
      ~ result           = (sensitive value)
        # (10 unchanged attributes hidden)
    }

Plan: 1 to add, 0 to change, 1 to destroy.
```

### Steps to Reproduce

1. With provider v2.2.1, do `terraform init`, `terraform apply`
2. Change to provider v3.4.0, do `terraform init -upgrade`, `terraform apply`

Note that going from v2.2.1 to v3.3.2 does not force replacement of the password. If I then update to v3.4.0 then the password is still not replaced. The password is only replaced with a jump upgrade straight from v2.2.1 to v3.4.0

If it's just impossible to support this jump upgrade then that's ok, but it would be ideal if this did work.

### How much impact is this issue causing?

Medium

### Logs

https://gist.github.com/blgm/e6f41c1209d1394a986344324bebc4ef

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

alexpoto:
[...]

PR #304, opened 2022-08-31
Fix upgrade v2.2.1 -> v3.4.0 forces replacement 

bendbennett:
References: #302 

This should address https://github.com/hashicorp/terraform-provider-random/issues/302 insofar as it fixes the issue when using the [config](https://github.com/hashicorp/terraform-provider-random/issues/302#issue-1357367589) supplied by [blgm](https://github.com/blgm) and the [config](https://github.com/hashicorp/terraform-provider-random/issues/302#issuecomment-1233045227) supplied by [alexpoto](https://github.com/alexpoto).


internal/provider/resource_password_test.go
```diff
@@ -506,6 +506,164 @@ func TestAccResourcePassword_Min(t *testing.T) {
 	})
 }
 
+// TestAccResourcePassword_UpgradeFromVersion2_2_1 requires that you are running an amd64 Terraform binary
+// if you are running this test locally on arm64 architecture otherwise you will see the following error:
+//
+// Error: Incompatible provider version
+//
+// Provider registry.terraform.io/hashicorp/random v2.2.1 does not have a
+// package available for your current platform ...
```
bflad, line 515: I think we can codify this logic into the test via:

```go
TestCase{
  PreCheck: func() {
    if runtime.GOOS == "darwin" && runtime.GOARCH == "arm64" {
      t.Skip("This test requires darwin/amd64 to download the old provider version. Setting TF_ACC_TERRAFORM_PATH to darwin/amd64 compatible Terraform binary can be used as a workaround.")
    }
  },
  // ... other fields
}
```

bendbennett: Thanks. Have incorporated.

bflad (approved):
Looks good to me 🚀

blgm:
Thank you @bendbennett and @bflad!

Issue #303, opened 2022-08-31
Breaking change of null keeper values between 3.3.2 and 3.4.0


danielvoros-form3:
### Terraform CLI and Provider Versions

Terraform v1.1.6
on linux_amd64
+ provider registry.terraform.io/hashicorp/random v3.4.0

### Terraform Configuration

```terraform
terraform {
  required_providers {
    random = {
      source = "hashicorp/random"
      version = "3.4.0"
    }
  }
}

provider "random" {
  # Configuration options
}

resource "random_integer" "rnd" {
    min = 1
    max = 100
    keepers = {
        "key" = null
    }
}
```


### Expected Behavior

Changing the provider version from `3.3.2` to `3.4.0` shouldn't force a replacement.

### Actual Behavior

Running a plan of the terraform code above with the provider version set to `3.3.2` and then with `3.4.0` forces a replacement. Apparently `null` keys have started to appear in the state with 3.4.0 while they weren't added to the state before.

### Steps to Reproduce

1. set provider version to `3.3.2`
2. `terraform init` + `terraform apply`
3. set provider version to `3.4.0`
4. `terraform init -upgrade`
5. `terraform plan`

Plan should be empty but it wants to recreate the resource instead:

```
Terraform will perform the following actions:                                                                                                                                                                                     
                                                                                                                                                                                                                                  
  # random_integer.rnd must be replaced                                                                                                                                                                                           
-/+ resource "random_integer" "rnd" {                                                                                                                                                                                             
      ~ id      = "21" -> (known after apply)                                                                                                                                                                                     
      + keepers = {                                                                                                                                                                                                               
          + "key" = null                                                                                                                                                                                                          
        } # forces replacement                                                                                                                                                                                                    
      ~ result  = 21 -> (known after apply)                                                                                                                                                                                       
        # (2 unchanged attributes hidden)                                                                                                                                                                                         
    }                                                                                                                                                                                                                             
                                                                                                                                                                                                                                  
Plan: 1 to add, 0 to change, 1 to destroy.
```

### How much impact is this issue causing?

High

### Logs

_No response_

### Additional Information

_No response_

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

anGie44:
Hey all 👋 I'm also seeing this when upgrading from version `3.3.2` to `3.4.1` 

my configuration:

```terraform
terraform {
  required_providers{
    random = {
      version = "3.4.1"
    }
  }
}

[...]

PR #305, opened 2022-09-01
Fix Breaking change of null keeper values between 3.3.2 and 3.4.0

bendbennett:
Closes: #303


bflad (approved):
Changes outside mine look good to me 🚀

PR #309, opened 2022-09-03
make referencing path to example generic

l-with:
since terraform provider random is mentioned as inspiring example for [terraform-plugin-docs](https://github.com/hashicorp/terraform-plugin-docs) the reference to the examples in the documentation templates is coded generic to avoid "copy-paste-forget-to-edit" failures


bflad (approved):
Looks good to me 🚀

Issue #306, opened 2022-09-02
random_password Showing Unexpected Plan 3.3.2 to 3.4.2


bflad:
### Terraform CLI and Provider Versions

Terraform v1.2.7
on darwin_arm64
+ provider registry.terraform.io/hashicorp/random v3.4.2

### Terraform Configuration

```terraform
terraform {
  required_providers {
    random = {
      source  = "hashicorp/random"
      version = "3.4.2"
      # version = "3.3.2"
    }
  }
}

resource "random_password" "test" {
  length = 20
}
```

### Expected Behavior

No plan difference.

### Actual Behavior

```console
$ terraform apply        
random_password.test: Refreshing state... [id=none]

Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  ~ update in-place

Terraform will perform the following actions:

  # random_password.test will be updated in-place
  ~ resource "random_password" "test" {
        id          = "none"
        # (12 unchanged attributes hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.
```

### Steps to Reproduce

1. `terraform apply`

### How much impact is this issue causing?

Medium

### Additional Information

May be related to small change in #305 where `override_special` was having issues with framework migration testing and the fix seemed to be adding a default value plan modifier of `types.String{Value: ""}`. Here's the state difference after the odd plan output:

```diff
4c4
<   "serial": 1,
---
>   "serial": 3,
28c28
<             "override_special": null,
---
>             "override_special": "",
33,34c33
<           "sensitive_attributes": [],
<           "private": "eyJzY2hlbWFfdmVyc2lvbiI6IjIifQ=="
---
>           "sensitive_attributes": []
```

Potentially removing `Computed` and removing the default value plan modifier can fix this.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

amh-mw:
Same behavior observed this morning for `random_string`, 3.4.1 -> 3.4.2.

fgm:
Same problem with random_string

SaravananGuru:
Any solution/suggestion on the same?  I'm also facing the same issue after upgrading the terraform azurerm version (v3.0.0).
can I safely ignore and what was the exact impact of this?  whether it will reset the random password string?

lucascantor:
@SaravananGuru pinning to `3.3.2` for the time being seems to resolve this issue for me:

```
terraform {
  required_version = ">= 1.2.2"
[...]

PR #310, opened 2022-09-05
Fix random_password Showing Unexpected Plan 3.3.2 to 3.4.2

bendbennett:
Closes: #306

### DIFF
--- a/internal/provider/resource_password.go
+++ b/internal/provider/resource_password.go
@@ -22,7 +22,7 @@ var _ provider.ResourceType = (*passwordResourceType)(nil)
 type passwordResourceType struct{}
 
 func (r *passwordResourceType) GetSchema(context.Context) (tfsdk.Schema, diag.Diagnostics) {
-	return passwordSchemaV2(), nil
+	return passwordSchemaV3(), nil
 }
 
 func (r *passwordResourceType) NewResource(_ context.Context, _ provider.Provider) (resource.Resource, diag.Diagnostics) {
@@ -38,7 +38,7 @@ var (
 type passwordResource struct{}
 
 func (r *passwordResource) Create(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) {
-	var plan passwordModelV2
+	var plan passwordModelV3
 
 	diags := req.Plan.Get(ctx, &plan)
 	resp.Diagnostics.Append(diags...)
@@ -65,7 +65,7 @@ func (r *passwordResource) Create(ctx context.Context, req resource.CreateReques
 		return
 	}
 
-	state := passwordModelV2{
+	state := passwordModelV3{
 		ID:              types.String{Value: "none"},
 		Keepers:         plan.Keepers,
 		Length:          types.Int64{Value: plan.Length.Value},
@@ -102,7 +102,7 @@ func (r *passwordResource) Read(ctx context.Context, req resource.ReadRequest, r
 
 // Update ensures the plan value is copied to the state to complete the update.
 func (r *passwordResource) Update(ctx context.Context, req resource.UpdateRequest, resp *resource.UpdateResponse) {
-	var model passwordModelV2
+	var model passwordModelV3
 
 	resp.Diagnostics.Append(req.Plan.Get(ctx, &model)...)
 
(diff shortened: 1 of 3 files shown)


internal/provider/resource_password.go
```diff
@@ -121,19 +121,21 @@ func (r *passwordResource) Delete(ctx context.Context, req resource.DeleteReques
...
+		Upper:           types.Bool{Value: true},
+		Lower:           types.Bool{Value: true},
+		Number:          types.Bool{Value: true},
+		Numeric:         types.Bool{Value: true},
+		MinSpecial:      types.Int64{Value: 0},
+		MinUpper:        types.Int64{Value: 0},
+		MinLower:        types.Int64{Value: 0},
+		MinNumeric:      types.Int64{Value: 0},
+		Keepers:         types.Map{Null: true},
+		OverrideSpecial: types.String{Null: true},
```
bflad, line 138: Do you happen to know what actually gets put into the state if these aren't set? I'm thinking it might be zero-values for the types, so empty map and empty string. 🙁 It'd be a bummer if provider developers explicitly had to set null for Terraform-only attributes on import. I've kind of wanted to support `types` pointers, e.g. `*types.String` in the reflection and potentially drop the `Null` field (and maybe further something like https://github.com/hashicorp/terraform-plugin-framework/issues/441 so you have to explicitly tell the framework that value is known vs unknown), but maybe that is more confusing.

internal/provider/resource_password.go
```diff
@@ -192,8 +199,20 @@ func upgradePasswordStateV0toV2(ctx context.Context, req resource.UpgradeStateRe
...
+
+	if len(passwordDataV0.Keepers.Elems) > 0 {
+		keepers = passwordDataV0.Keepers
+	}
+
+	overrideSpecial := types.String{Null: true}
+
+	if len(passwordDataV0.OverrideSpecial.Value) > 0 {
+		overrideSpecial = passwordDataV0.OverrideSpecial
+	}
```
bflad, line 212: I _want_ to do this but I believe it would unfortunately cause practitioner plan differences if their configuration did happen to be:

```go
resource "random_password" "example" {
  # ... other attributes ...
  override_special = ""
}
```

Which is perfectly valid and may be coming from something like a module variable declaration:

```terraform
variable "password_special_characters" {
  default     = ""
  description = "Set this to override the special characters that can be included in the password. Defaults to !@#$%&*()-_=+[]{}<>:?"
  type        = string
}

resource "random_password" "example" {
  # ... other attributes ...
  override_special = var.password_special_characters
}
```

For example, the next plan after state upgrade I believe Terraform would show the below (or potentially would show that there's an in-place update, but not specifically this line because it might be silenced due to legacy SDK plan handling):

```
# resource random_password.test will be updated in-place
~ resource "random_password" "test" {
  ~ override_special: null => ""
}
```

internal/provider/resource_password.go
bendbennett: If `Keepers` and `OverrideSpecial` are not set in the `ImportState` func, then as you thought, we end up with the zero value for the types in the state. 

```
"keepers": {},
...
"override_special": "",
```

This is problematic as the next _terraform apply_ results in a plan difference.

internal/provider/resource_password.go
bendbennett: You are correct. Adding a state upgrader to do the following:

```go
	keepers := types.Map{Null: true, ElemType: types.StringType}

	if len(passwordDataV2.Keepers.Elems) > 0 {
		keepers = passwordDataV2.Keepers
	}

	overrideSpecial := types.String{Null: true}

	if len(passwordDataV2.OverrideSpecial.Value) > 0 {
		overrideSpecial = passwordDataV2.OverrideSpecial
	}
```

Will result in practitioner plan differences if they used `"override_special": ""` and/or `"keepers": {}` in their configuration.

If `"override_special": ""` is used in the config to create the resource (e.g., using `v3.4.2`) and then _terraform apply_ is run with this branch the following is seen:
```
  # random_password.test must be replaced
-/+ resource "random_password" "test" {
      ~ bcrypt_hash = (sensitive value)
      ~ id          = "none" -> (known after apply)
      ~ result      = (sensitive value)
        # (10 unchanged attributes hidden)
    }
```


If `"keepers": {}` is used in the config to create the resource (e.g., using `v3.4.2`) and then _terraform apply_ is run with this branch the following is seen:
```
  # random_password.test will be updated in-place
  ~ resource "random_password" "test" {
        id          = "none"
      + keepers     = {}
        # (12 unchanged attributes hidden)
    }
```

bendbennett:
Closing as superseded by #312

PR #313, opened 2022-09-06
Implement State Upgrader to Set Defaults for Attributes Assigned Null During Import

bendbennett:
In the discussion contained in [Upgrade v2.2.1 -> v3.4.0 forces replacement](https://github.com/hashicorp/terraform-provider-random/issues/302#top) there were several comments indicating that a forced replacement was happening under some circumstances:

- https://github.com/hashicorp/terraform-provider-random/issues/302#issuecomment-1233159123
- https://github.com/hashicorp/terraform-provider-random/issues/302#issuecomment-1233263617
- https://github.com/hashicorp/terraform-provider-random/issues/302#issuecomment-1235611077

The attributes in question usually have a default value of `0` assigned (i.e., `min_lower`, `min_upper`, `min_numeric` and `min_special`) or `true` assigned (i.e., `lower`, `upper`, `number`, `numeric` and `special`) when no value for them is specified in the configuration.

However, under some circumstances, the attributes described can end up with a value of `null` assigned to them in the state which results in a forced update when _terraform apply_ is run.

One of the circumstances that can give rise to the storage of null values for these attributes is using _terraform import_ with a version of the provider that is `v3.3.1` or earlier. Under these circumstances, all of the attributes mentioned, along with `length` are assigned a value of `null` in state.

This PR aims to address this issue by implementing a state upgrader which interrogates each of the aforementioned attributes and assigns it a default value if the value currently held for the attribute in state is `null`.

### DIFF
--- a/internal/provider/provider_test.go
+++ b/internal/provider/provider_test.go
@@ -22,6 +22,15 @@ func providerVersion221() map[string]resource.ExternalProvider {
 	}
 }
 
+func providerVersion313() map[string]resource.ExternalProvider {
+	return map[string]resource.ExternalProvider{
+		"random": {
+			VersionConstraint: "3.1.3",
+			Source:            "hashicorp/random",
+		},
+	}
+}
+
 func providerVersion320() map[string]resource.ExternalProvider {
 	return map[string]resource.ExternalProvider{
 		"random": {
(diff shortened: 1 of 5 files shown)


bendbennett:
I've marked this as ready for review as the code changes should be complete.

Checks will fail until `terraform-plugin-sdk` `v2.22.0` containing `ImportStatePersist` is released and this branch is updated to use that tag.

bflad (approved):
Looks good to me 🚀

Issue #377, opened 2023-03-03
Bump Expected Minimum Go Version to 1.19


bendbennett:
### Terraform CLI and Provider Versions

random v3.4.3

### Use Cases or Problem Statement

Following the Go [support policy](https://golang.org/doc/devel/release.html#policy) and given the ecosystem availability of the latest Go minor version, it's time to upgrade. This will ensure that this project can use recent improvements to the Go runtime, standard library functionality, and continue to receive security updates

### Proposal

* Run the following commands to upgrade the Go module files and automatically fix outdated Go code:

```shell
go mod edit -go=1.19
go mod tidy
go fix ./...
```

* Ensure any GitHub Actions workflows (`.github/workflows/*.yml`) use 1.20 in place of any 1.19 and 1.19 in place of any 1.18 or earlier
* Ensure the README or any Contributing documentation notes the Go 1.19 expected minimum
* (Not applicable to all projects) Ensure the `.go-version` is at least 1.19 or later

### How much impact is this issue causing?

Low

### Additional Information

- https://github.com/hashicorp/terraform-plugin-framework/#go-compatibility
- https://formulae.brew.sh/formula/go
- https://github.com/microsoft/winget-pkgs/pull/95835
- https://community.chocolatey.org/packages/golang
- https://packages.ubuntu.com/lunar/golang-1.20

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

faarshad:
Hi @bendbennett, thank you for fixing this. I see https://github.com/hashicorp/terraform-provider-random/pull/378 is merged to main but any idea on when this would be released?

It will fix all these security bugs:
     1	CVE-2022-32190|go|.terraform.d/plugins/registry.terraform.io/hashicorp/random/3.4.3/linux_amd64/terraform-provider-random_v3.4.3_x5|1.18.5|fixed in 1.18.6|7.5|high
     2	CVE-2022-27664|go|.terraform.d/plugins/registry.terraform.io/hashicorp/random/3.4.3/linux_amd64/terraform-provider-random_v3.4.3_x5|1.18.5|fixed in 1.19.1, 1.18.6|7.5|high
     3	CVE-2022-2879|go|.terraform.d/plugins/registry.terraform.io/hashicorp/random/3.4.3/linux_amd64/terraform-provider-random_v3.4.3_x5|1.18.5|fixed in 1.19.2, 1.18.7|7.5|high
[...]

PR #378, opened 2023-03-03
Updating to Go 1.19

bendbennett:
Closes: #377


bflad (approved):
Looks good to me 🚀  (this provider is only unit testing with the `go.mod` version, which is fine for provider codebases)

Issue #396, opened 2023-04-12
Unable to generate passwords with >72 characters


henryde:
### Terraform CLI and Provider Versions

terraform: 1.4.2
provider: 3.5.0

### Terraform Configuration

```terraform
resource "random_password" "test" {
  length  = 73
}
```

### Expected Behavior

generating longer passwords was possible with 3.4.3

### Actual Behavior

```
╷
│ Error: Hash Generation Error
│
│   with random_password.test,
│   on test.tf line 1, in resource "random_password" "test":
│    1: resource "random_password" "test" {
│
│ While attempting to generate a hash from the password an error occurred.
│
│ Verify that the state contains a populated 'result' field, using 'terraform state show', and retry the operation
│
│ Original Error: bcrypt: password length exceeds 72 bytes
╵
```

### Steps to Reproduce

1. `terraform apply`

### How much impact is this issue causing?

Medium

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

ashhammond:
We are also seeing the exact same issue with the latest version of the provider. Rolling back and testing as suggested above.

ashhammond:
As a workaround we have now pinned the random provider to 3.4.3 and functionality is restored.

austinvalle:
Hi there @henryde 👋🏻 , thanks for submitting the issue and sorry you ran into this bug.

We've merged a fix for this bug with #397 and released `v3.5.1` of the random provider with this bug fix. You should see that provider version become available in the next hour or so.

ashhammond:
thanks @austinvalle for the quick fix.

PR #397, opened 2023-04-12
Prevent hash error `bcrypt.ErrPasswordTooLong` for passwords  > 72 bytes in length

bendbennett:
Closes: #396 

This PR replicates the behaviour of the `bcrypt` package in versions of [golang.org/x/crypto](https://pkg.go.dev/golang.org/x/crypto/bcrypt?tab=versions) prior to `v0.5.0`.

### DIFF
--- a/.changes/unreleased/BUG FIXES-20230412-132501.yaml
+++ b/.changes/unreleased/BUG FIXES-20230412-132501.yaml
@@ -1,5 +1,5 @@
 kind: BUG FIXES
-body: 'resource/random_password: The `generateHash()` functions has been modified
+body: 'resource/random_password: The `generateHash()` function has been modified
   to truncate strings to 72 bytes'
 time: 2023-04-12T13:25:01.113462+01:00
 custom:
--- a/.changes/unreleased/BUG FIXES-20230412-132501.yaml
+++ b/.changes/unreleased/BUG FIXES-20230412-132501.yaml
@@ -1,6 +1,6 @@
 kind: BUG FIXES
-body: 'resource/random_password: The `generateHash()` function has been modified
-  to truncate strings to 72 bytes'
+body: 'resource/random_password: Prevent error with `bcrypt` by truncating the bytes
+  that are hashed to a maximum length of 72'
 time: 2023-04-12T13:25:01.113462+01:00
 custom:
   Issue: "397"
--- a/.changes/unreleased/BUG FIXES-20230412-132501.yaml
+++ b/.changes/unreleased/BUG FIXES-20230412-132501.yaml
@@ -1,0 +1,6 @@
+kind: BUG FIXES
+body: 'resource/random_password: The `generateHash()` functions has been modified
+  to truncate strings to 72 bytes'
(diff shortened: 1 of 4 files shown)


.changes/unreleased/BUG FIXES-20230412-132501.yaml
```diff
@@ -0,0 +1,6 @@
+kind: BUG FIXES
+body: 'resource/random_password: The `generateHash()` functions has been modified
```
bookshelfdave, line 2: typo: functions -> function

bookshelfdave (approved):

.changes/unreleased/BUG FIXES-20230412-132501.yaml
bflad: Open question: will practitioners know what "generateHash()" is? I'd suggest tailoring the message for the behavior they will receive, which I believe is that we are preventing an error with `bcrypt` by truncating that hash to 72 bytes as it was previously. I'm sure that may raise further questions though.

austinvalle (approved):
lgtm 👍🏻, can we also add a note about this behavior to the description of the output `bcrypt_hash`, so it shows in the docs.

internal/provider/resource_password.go
```diff
@@ -535,8 +535,21 @@ func upgradePasswordStateV2toV3(ctx context.Context, req resource.UpgradeStateRe
 	resp.Diagnostics.Append(resp.State.Set(ctx, passwordDataV3)...)
 }
 
+// generateHash truncates strings that are longer than 72 bytes in
+// order to avoid the error returned from bcrypt.GenerateFromPassword
+// in versions v0.5.0 and above (refer to
+// https://pkg.go.dev/golang.org/x/crypto/bcrypt?tab=versions).
+// This is to preserve backwards compatibility and will be removed in
+// a subsequent version of the provider.
```
bflad, line 543: Is the plan to validate password length to a maximum of 72 bytes in a future version or should this comment be removed?

bflad (approved):
Given the prior behavior and current, more unfortunate regression behavior, this looks good to me.

internal/provider/resource_password.go
bendbennett: I was thinking in the next major version release we could generate an error if the password length exceeds 72 bytes. If that sounds ok, I'll create a follow-up issue and link it in the code comment.

bflad: Please feel free to raise a proposal, however I'm a little worried that may be an unpopular proposal just to appease a specific hashing algorithm behavior that not all affected practitioners (anyone generating long passwords) may need from the resource. I think noting the limitation in that hash description will either be good enough for the purposes of this provider or another proposal could be made to move that hashing algorithm to another provider (potentially in a provider-defined function once available) and drop the "problematic" attribute as out of scope for this provider.

austinvalle: I removed the statement of removal from the comment and we can determine in a later issue if that's something we want to do ⚡

Issue #436, opened 2023-07-24
Documentation: random_shuffle output is already list


iancward:
### Terraform CLI and Provider Versions

N/A

### Terraform Configuration

```terraform
resource "random_shuffle" "az" {
  input        = ["us-west-1a", "us-west-1c", "us-west-1d", "us-west-1e"]
  result_count = 2
}

resource "aws_elb" "example" {
  # Place the ELB in any two of the given availability zones, selected
  # at random.
  availability_zones = [random_shuffle.az.result]

  # ... and other aws_elb arguments ...
}
```

### Expected Behavior

`random_shuffle` output is already a list of string, so it does not need to be put into another list when used, for example, in `availability_zones` argument of `aws_elb`.

### Actual Behavior

N'A

### Steps to Reproduce

N/A

### How much impact is this issue causing?

Low

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

bendbennett:
Hi @iancward 👋 

Thank you for raising this issue.
I've opened a [PR](https://github.com/hashicorp/terraform-provider-random/pull/438) to amend the documetation.

PR #438, opened 2023-07-25
Amending documentation to reflect that the result from random_shuffle is a list

bendbennett:
Closes: #436


bflad (approved):
Looks good to me 🚀 If I had to venture a wild guess, the extra square braces might have been helpful for very old versions of Terraform (pre-0.12 days).

Issue #549, opened 2024-04-09
Panic on `random_string` when all properties set to `false`


guineveresaenger:
### Terraform CLI and Provider Versions

Terraform v1.5.7
on darwin_arm64
+ provider registry.terraform.io/hashicorp/random v3.6.0

### Terraform Configuration

```terraform
terraform {
  required_providers {
    random = {
      version = "3.6.0"
    }
  }
}

provider "random" {

}

# Create a random string resource
resource "random_string" "my_string" {
  length  = 16
  special = false
  upper   = false
  number  = false
  lower   = false
}

# Output the generated random string
output "random_string_result" {
  value = random_string.my_string.result
}
```

### Expected Behavior

I should see an error telling me my configuration is nonsensical.

### Actual Behavior

I got a panic.

### Steps to Reproduce

1. `terraform apply`
2. See panic stack trace:

```
Stack trace from the terraform-provider-random_v3.6.0_x5 plugin:

panic: crypto/rand: argument to Int is <= 0

goroutine 26 [running]:
crypto/rand.Int({0x102d2e600?, 0x14000142380?}, 0xffffffffffffffff?)
	crypto/rand/util.go:64 +0x29c
github.com/terraform-providers/terraform-provider-random/internal/random.generateRandomBytes(0x14000182c08, 0x10)
	github.com/terraform-providers/terraform-provider-random/internal/random/string.go:90 +0xc8
github.com/terraform-providers/terraform-provider-random/internal/random.CreateString({0x10, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, {0x0, ...}})
	github.com/terraform-providers/terraform-provider-random/internal/random/string.go:67 +0x35c
github.com/terraform-providers/terraform-provider-random/internal/provider.(*stringResource).Create(0xd?, {0x102d34738, 0x14000354cf0}, {{{{0x102d38960, 0x14000355ce0}, {0x102c585e0, 0x14000355b60}}, {0x102d39b18, 0x14000136d20}}, {{{0x102d38960, ...}, ...}, ...}, ...}, ...)
	github.com/terraform-providers/terraform-provider-random/internal/provider/resource_string.go:71 +0x2c8
github.com/hashicorp/terraform-plugin-framework/internal/fwserver.(*Server).CreateResource(0x1400015d600, {0x102d34738, 0x14000354cf0}, 0x14000183320, 0x140001832c0)
[...]

PR #550, opened 2024-04-10
Error handling: Check that char set is not empty

guineveresaenger:
fix: return error instead of panic when charset winds up empty. Add test.

Fixes https://github.com/hashicorp/terraform-provider-random/issues/549.


guineveresaenger:
Closing in favor of #551.

PR #551, opened 2024-04-10
Prevent panic on `random_string` and `random_password` when character set is empty

bendbennett:
Closes: #549

### DIFF
--- a/internal/provider/resource_password.go
+++ b/internal/provider/resource_password.go
@@ -26,6 +26,7 @@ import (
 	mapplanmodifiers "github.com/terraform-providers/terraform-provider-random/internal/planmodifiers/map"
 	stringplanmodifiers "github.com/terraform-providers/terraform-provider-random/internal/planmodifiers/string"
 	"github.com/terraform-providers/terraform-provider-random/internal/random"
+	"github.com/terraform-providers/terraform-provider-random/internal/validators"
 )
 
 var (
@@ -630,6 +631,13 @@ func passwordSchemaV3() schema.Schema {
 					boolplanmodifier.RequiresReplace(),
 				},
 				DeprecationMessage: "**NOTE**: This is deprecated, use `numeric` instead.",
+				Validators: []validator.Bool{
+					validators.AtLeastOneOfTrue(
+						path.MatchRoot("special"),
+						path.MatchRoot("upper"),
+						path.MatchRoot("lower"),
+					),
+				},
 			},
 
 			"numeric": schema.BoolAttribute{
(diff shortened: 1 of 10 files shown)


internal/validators/boolvalidator.go
```diff
@@ -0,0 +1,116 @@
...
+
+	"github.com/hashicorp/terraform-plugin-framework-validators/helpers/validatordiag"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
+	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+// AtLeastOneOfTrueValidator is the underlying struct implementing AtLeastOneOfTrue.
```
SBGoods, line 18: We can probably accomplish the same thing with some sort of combination of the built-in `any()`, `all()`, `atLeastOneOf()` bool validators with an `equals()` validator, but I think that your approach is more readable.

SBGoods (approved):
Just one comment about documentation but otherwise LGTM!

internal/provider/resource_string.go
```diff
@@ -442,6 +450,13 @@ func stringSchemaV3() schema.Schema {
 					boolplanmodifiers.NumberNumericAttributePlanModifier(),
 					boolplanmodifier.RequiresReplace(),
 				},
+				Validators: []validator.Bool{
+					validators.AtLeastOneOfTrue(
```
bflad, line 454: Nota Bene: Definitely don't need to change this, but another consideration here might be to introduce this multiple attribute validation at the resource level as a ConfigValidator and specify all 4 attributes in there. Maybe its just me but I tend to get a little confused reading attribute-based validators (remembering "oh yeah, it includes this attribute too") and whether that attribute-based validation should be added to all the relevant attributes (confusingly, no, since it will generate multiple similar diagnostics). 😅

internal/provider/resource_password.go
```diff
@@ -630,6 +631,13 @@ func passwordSchemaV3() schema.Schema {
...
+				Validators: []validator.Bool{
+					validators.AtLeastOneOfTrue(
+						path.MatchRoot("special"),
+						path.MatchRoot("upper"),
+						path.MatchRoot("lower"),
+					),
+				},
 			},
 
 			"numeric": schema.BoolAttribute{
```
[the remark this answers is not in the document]
bendbennett: Good shout. I've updated the description for `number` and `numeric` for both `resource_password` and `resource_string` and regenerated the docs.

internal/provider/resource_string.go
bendbennett: Happy to refactor along these lines, but if you're ok with the current implementation could you approve the PR?

internal/validators/boolvalidator.go
bendbennett: Agreed. But like you, this seemed more readable to me.

bflad (approved):

PR #554, opened 2024-04-16
workflows: Remove wildcard suffix from Terraform workflow call

austinvalle:
Can be merged after -> https://github.com/hashicorp/terraform-devex-repos/pull/69


SBGoods (approved):

austinvalle:
Action failure isn't related to the workflow changes here, although we'll need to investigate what's going on with the incorrect plan diff

PR #558, opened 2024-04-19
resource/random_password: Fix testing for Terraform 1.8 non-refresh handling

bflad:
Reference: https://developer.hashicorp.com/terraform/language/v1.8.x/upgrade-guides#possible-spurious-changes-when-refreshing

Previously the `TestAccResourcePassword_Import_FromVersion3_1_3` acceptance test used `PlanOnly: true` which ran a non-refresh plan as its first step. After Terraform 1.8, additional attribute sensitivity metadata is saved into the state, which can cause a non-refresh plan difference and break the acceptance test:

```
=== CONT  TestAccResourcePassword_Import_FromVersion3_1_3
    resource_password_test.go:357: Step 2/3 error: The non-refresh plan was not empty.
        stdout:

        Terraform used the selected providers to generate the following execution
        plan. Resource actions are indicated with the following symbols:
          ~ update in-place

        Terraform will perform the following actions:

          # random_password.test will be updated in-place
          ~ resource "random_password" "test" {
                id          = "none"
                # (12 unchanged attributes hidden)
            }

        Plan: 0 to add, 1 to change, 0 to destroy.
    panic.go:626: Error retrieving state, there may be dangling resources: exit status 1
        Failed to marshal state to json: schema version 0 for random_password.test in state does not match version 3 from the provider
--- FAIL: TestAccResourcePassword_Import_FromVersion3_1_3 (1.71s)
```

Adjusting this acceptance test to replace `PlanOnly: true` with `plancheck.ExpectNoChanges()` covers the same desired behavior along with the more-common practitioner behavior of refresh plans rather than non-refresh plans after provider upgrades.

### DIFF
--- a/internal/provider/resource_password_test.go
+++ b/internal/provider/resource_password_test.go
@@ -17,6 +17,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
 	"github.com/hashicorp/terraform-plugin-go/tftypes"
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/plancheck"
 	"github.com/hashicorp/terraform-plugin-testing/terraform"
 	"golang.org/x/crypto/bcrypt"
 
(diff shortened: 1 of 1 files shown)


austinvalle (approved):
🚀

PR #599, opened 2024-08-07
testing: Update acceptance tests to use new `statecheck` package and value comparers

austinvalle:
Ref: https://github.com/hashicorp/terraform-plugin-testing/issues/295, https://github.com/hashicorp/terraform-plugin-testing/issues/266

With this refactor, all of the assertions should be equivalent except for a few where I made the test assertions more restrictive/exact. Let me know if I missed anything!


go.mod
```diff
@@ -8,10 +8,11 @@ require (
 	github.com/dustinkirkland/golang-petname v0.0.0-20240428194347-eebcea082ee0
 	github.com/google/go-cmp v0.6.0
 	github.com/hashicorp/go-uuid v1.0.3
+	github.com/hashicorp/terraform-json v0.22.1
 	github.com/hashicorp/terraform-plugin-framework v1.11.0
 	github.com/hashicorp/terraform-plugin-framework-validators v0.13.0
 	github.com/hashicorp/terraform-plugin-go v0.23.0
-	github.com/hashicorp/terraform-plugin-testing v1.9.0
+	github.com/hashicorp/terraform-plugin-testing v1.9.1-0.20240807155751-bcdd810e0d34
```
austinvalle, line 15: TODO: Replace with released `terraform-plugin-testing` pkg version

internal/provider/resource_password_test.go
```diff
@@ -365,6 +382,8 @@ func TestAccResourcePassword_Import_FromVersion3_1_3(t *testing.T) {
 				ImportState:        true,
 				ImportStateId:      "Z=:cbrJE?Ltg",
 				ImportStatePersist: true,
+				// TODO: Import state checks haven't been implemented in terraform-plugin-testing yet, so can't use value comparers for now
+				// TODO: Create import state check issue in terraform-plugin-testing
```
austinvalle, line 386: We'll need to introduce support for state checks in import to update these tests, perhaps a miss on our original implementation, but I don't think it should be too difficult to introduce 🤞🏻

internal/provider/resource_password_test.go
```diff
@@ -559,13 +590,17 @@ func TestAccResourcePassword_StateUpgradeV0toV3(t *testing.T) {
 			configDuringUpgrade: `resource "random_password" "default" {
 						length = 12
 					}`,
-			beforeStateUpgrade: []resource.TestCheckFunc{
-				resource.TestCheckResourceAttr("random_password.default", "number", "true"),
-				resource.TestCheckNoResourceAttr("random_password.default", "numeric"),
+			beforeUpgradeStateChecks: []statecheck.StateCheck{
+				statecheck.ExpectKnownValue("random_password.default", tfjsonpath.New("number"), knownvalue.Bool(true)),
+				randomtest.ExpectNoAttribute("random_password.default", tfjsonpath.New("numeric")),
```
austinvalle, line 595: A fun (and expected) quirk of our new state checks are that they will recognize if an attribute doesn't exist and fail, so certain usages of `resource.TestCheckNoResourceAttr` will not be able to be replaced by `statecheck.ExpectKnownValue` with a `knownvalue.Null()`

internal/provider/resource_password_test.go
austinvalle: Opened an issue and will update the TODO comments: https://github.com/hashicorp/terraform-plugin-testing/issues/365

SBGoods (approved):

Issue #630, opened 2024-12-11
random_string: add case-insensitive rejection string


ari-becker:
### Terraform CLI and Provider Versions

terraform 1.9.4
random: 3.6.3

### Use Cases or Problem Statement

We rely upon `random_string` to generate resource identifiers for Azure resources.

Today, in a particular case, it generated `xbox7zw6jugefzbnlvi8dl1z`.

This lead to the following error:
```
unexpected status 400 (400 Bad Request) with error: ReservedResourceName: The resource name 'xbox7zw6jugefzbnlvi8dl1z' or a part of the name is a trademarked or reserved word
```

So the generated string ran afoul of https://learn.microsoft.com/en-us/azure/azure-resource-manager/troubleshooting/error-reserved-resource-name and caused the Terraform apply to error out.

### Proposal

Some way of configuring the `random_string` provider with the Azure-provided list of reserved/trademarked strings in order to guarantee that the random provider doesn't provide random strings that violate this limitation.

### How much impact is this issue causing?

Medium

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

ramonvermeulen:
Would adding two properties:
* `exclusion_list` List of String
* `exclusion_list_case_sensitive` Boolean (default false)

Be a good solution to this issue? Ability to add an exclusion list either being case sensitive or case insensitive. When a random string gets generated and matches the exclusion list, it will regenerate until it has a string that does not match any of the strings in the exclusion list. If so, I'm willing to create a PR for this feature.

ari-becker:
> Would adding two properties:
> * `exclusions` Set of String
> * `exclusions_case_sensitive` Boolean (default false)
> 
> Be a good solution to this issue? Ability to add an exclusion set either being case sensitive or case insensitive. When a random string gets generated and matches the exclusion set, it will regenerate until it has a string that does not match any of the strings in the exclusion set. Another idea could be instead of a Set of Strings to accept a Set of Regex patterns, if the generated string matches the pattern it will regenerate a new one?
> 
> If so, I'm willing to create a PR for this feature.
> 
> EDIT: opened draft PR.

All of the above would be helpful :) thanks

austinvalle:
Hey there @ari-becker 👋🏻 , thanks for raising the issue and sorry you're running into trouble here.

Before looking to design a solution for this, I'd be interested in hearing:
- If you run into this issue often or if it was a one-time problem.
- What your workaround looked like when you ran into this

[...]

PR #645, opened 2025-01-27
feat: initial draft implementation for string exclusions

ramonvermeulen:
Added a draft implementation for issue: https://github.com/hashicorp/terraform-provider-random/issues/630

Adds two properties to the `random_string` as well as `random_password` resource:

* `exclusions` Set of String
* `exclusions_case_sensitive` Boolean (default false)

The goal is to provide a set of exclusions for the generated string (or password), if the generated string matches any of the exclusions it will regenerate the string until it has no match. If preferred I can change the exclusions to a configurable set of Regex patterns (might offer more flexibility).

Because it is my first time contributing to the terraform-provider-random, I'd love to first get a review on the following questions before I proceed with the implementation (e.g. going deep on testing and finalizing the implementation).

1. Is this functionality that is desired in the terraform random provider? or would it be considered out-of-scope for the provider.
2. Should `exclusions` be changed to a set of Regex instead of a plain string, I think it might offer more flexibility.
3. Would this implementation introduce a breaking change? As far I can see it is added functionality, e.g. adding configurable arguments and not modifying or removing existing arguments, so could be put into a minor release. However, I don't know the terraform-provider-random good enough yet to overview all the implications.

### DIFF
--- a/internal/planmodifiers/bool/boolplanmodifiers.go
+++ b/internal/planmodifiers/bool/boolplanmodifiers.go
@@ -5,8 +5,8 @@ package boolplanmodifiers
 
 import (
 	"context"
-
 	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/boolplanmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/types"
 )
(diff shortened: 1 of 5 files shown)


ramonvermeulen:
Hi @austinvalle,

I’ve been following the activity in the repository and noticed that you've been one of the most active maintainers over the past year. I was wondering if you might have some time to review my proposal above. I’d really appreciate your feedback.

austinvalle:
Hey @ramonvermeulen 👋🏻 , thanks for the PR!

I do have some things I'd like to hear more about with the use-case before proceeding with the PR, which I've added in a comment over on the issue if you'd like to add your thoughts: https://github.com/hashicorp/terraform-provider-random/issues/630#issuecomment-2657865487

PR #706, opened 2025-06-26
Add resources for uuid v4 and v7

### LINKED ISSUE #402: Improve uuid generated by making it valid

Terraform v1.4.5 - on darwin_amd64

provider 3.5.1

Currently random_uuid doesn't generate uuid's that follow the [RFC 4122](http://tools.ietf.org/html/rfc4122) and can be validated against the same RFC.

Some applications we interact with do uuid's validation which is causing some problems with terraform automation of those uuid's across multiple services.

### Proposal

migrate from go-uuid to github.com/google/uuid library

Medium

### Additional Information

_No response_

### Code of Conduct

### LINKED ISSUE #511: Support UUIDv7

N/A

Current lib doesn't allow the generation of UUID v7 format.
https://datatracker.ietf.org/doc/draft-ietf-uuidrev-rfc4122bis/14/

### Proposal

Implement ability to generate UUIDv7 unmbers.

Medium

### Additional Information

The new v7 style UUIDs have some very useful properties, and being able to generate and use them in terraform will allow terraform to be used in cases where those numbers are needed

### Code of Conduct

masongup-mdsol:
## Related Issue

Fixes #402 
Fixes #511 

## Description

The UUIDs generated by the existing `random_uuid` resource are not valid UUIDs per RFC 4122, so they cause problems when a user attempts to use them in a system which expects UUIDs to be a specific version. Per the discussion on #402, I have done this by cloning the existing uuid resource, renaming it as uuid4 and uuid7 respectively, and using the [Google Go UUID library](https://github.com/google/uuid) instead with the appropriate methods to generate valid UUIDs of the appropriate versions. This way, users of the existing `uuid` resource will continue to have the same behavior without making any changes. Users who need one of these UUID versions can use the specific resource for that so that it is clear to any readers of the code what version is being used and that that version is being used intentionally.

## Rollback Plan

- [ ] If a change needs to be reverted, we will roll out an update to the code within 7 days.

## Changes to Security Controls

Given that this is random data, there are no security controls involved.

### DIFF
--- a/.changes/unreleased/ENHANCEMENTS-20250918-121503.yaml
+++ b/.changes/unreleased/ENHANCEMENTS-20250918-121503.yaml
@@ -1,0 +1,5 @@
+kind: ENHANCEMENTS
+body: Add uuid4 and uuid7 resources to generate valid random uuids of the appropriate version
+time: 2025-09-18T12:15:03.335115-04:00
+custom:
+    Issue: "402"


masongup-mdsol:
Ah okay, I thought I had signed it already, but I guess I didn't finish the workflow properly before. I think it should be good now.

chrisjaimon2012:
@rainkwan can you take a look at this when you get some time?

rainkwan:
Hi @masongup-mdsol thank you for your contribution! 😄

This looks pretty reasonable, I wanted to ask if you could create a [changelog entry](https://github.com/hashicorp/terraform-provider-http/blob/main/.github/CONTRIBUTING.md#changie-automation-tool) for this as an enhancement. And also if you can run `make generate` so that the CI passes.

Thank you!

Issue #741, opened 2025-12-11
random_password & random_string are unable to deal with multibyte utf-8 character


kstiehl:
### Terraform CLI and Provider Versions

```
...
+ provider registry.opentofu.org/hashicorp/random v3.7.2
...
```

### Terraform Configuration

```terraform
resource "random_password" "backup" {
  length           = 32
  min_special      = 5
  override_special = "^°!§$%/()=?+#-.,;:~*@{}_&üäö"
}
```

### Expected Behavior

The resource should generate a valid 32-character password containing exactly 5 special characters randomly selected from the provided override_special string. Since Terraform uses UTF-8 encoding (inherited from Go), the password generator should correctly handle multi-byte UTF-8 characters such as German umlauts (ä, ü, ö, ß) and other European characters (€, ñ, ç, °) without corruption.

### Actual Behavior

Sometimes a butchered passowrd is provided which is not even a valid string

### Steps to Reproduce

1. `terraform apply`

### How much impact is this issue causing?

Medium

### Additional Information

I created a PR for this https://github.com/hashicorp/terraform-provider-random/pull/742

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

kstiehl:
https://github.com/hashicorp/terraform-provider-random/pull/742

PR #742, opened 2025-12-11
fix(random_password & random_string): CreateString is not generating valid strings 

kstiehl:
## Related Issue

Fixes: https://github.com/hashicorp/terraform-provider-random/issues/741

## Problem Description

The `random_password` resource in Terraform currently supports custom special characters through user configuration. While Terraform strings are UTF-8 encoded (inherited from Go's UTF-8 string handling), the `CreateString` function incorrectly processes multi-byte UTF-8 characters when generating passwords.

## Technical Details

When users specify special characters that are commonly used in European contexts—such as:
- German umlauts: `ä`, `ü`, `ö`, `ß`
- Currency symbols: `€`
- Accented characters: `ñ`, `ç`
- Other symbols: `°`

The `CreateString` function corrupts these characters because it operates at the byte level rather than the character level. Since these are all multi-byte UTF-8 characters (requiring 2-4 bytes each), the function's random byte selection breaks the character encoding by potentially:
- Selecting only partial bytes from a multi-byte sequence
- Creating invalid UTF-8 sequences
- Producing garbled output instead of the intended characters

## Root Cause

The fundamental issue is that `CreateString` treats the input string as a sequence of bytes rather than a sequence of UTF-8 characters, leading to malformed character sequences when multi-byte characters are included in the allowed character set.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,6 +12,7 @@ require (
 	github.com/hashicorp/terraform-plugin-framework-validators v0.19.0
 	github.com/hashicorp/terraform-plugin-go v0.29.0
 	github.com/hashicorp/terraform-plugin-testing v1.14.0
+    github.com/stretchr/testify v1.10.0
 	golang.org/x/crypto v0.48.0
 )
 


kstiehl:
Hi @rainkwan,
Sorry that it took so long to come back to this issue.

- I added the copyright header the linter was complaining about.
- The changelog entry was added.
- I added an acceptance test next to the existing ones. 
  - I fixed the StringExactLength function (since it is more correct now, and it does not break any other test or assumptions of the code)

PR #745, opened 2025-12-19
random_string: Generate value at plan-time

austinvalle:
## Related Issues

Ref #121
Ref #487

Based off WIP protocol changes made in:
- https://github.com/hashicorp/terraform/pull/37986
- https://github.com/hashicorp/terraform-plugin-go/tree/av/planned-private
- https://github.com/hashicorp/terraform-plugin-framework/tree/av/planned-private

## Description

This PR enhances the `random_string` resource to generate it's value at plan time if supported by the client (currently no released versions of Terraform support this). It does this by persisting data in private state between the initial and final plan (if necessary) so that the random value is generated only once. If the value can't be persisted by the client then it behaves as it does today (unknown value at plan time, generated during apply).

This PR only enhances `random_string`, but most likely all of the other random resources should receive the same enhancement in follow-up PRs, for example #487 .

This PR is also just a quick implementation and doesn't have any tests 😆, which should be added before opening for review. It also requires TF core, `terraform-plugin-go`, and `terraform-plugin-framework` changes to be released before it can be merged. 

### Current behavior

https://github.com/user-attachments/assets/3a70e574-17a7-4f18-9f39-a8ba8e5f15eb

### New behavior

https://github.com/user-attachments/assets/db170027-764b-4e6d-be8a-76ffa1f38bba

## Rollback Plan

- [x] If a change needs to be reverted, we will roll out an update to the code within 7 days.

## Changes to Security Controls

No


austinvalle:
I'm going to close this PR as I'm no longer working on the team that owns this provider so it's unlikely this PR will get merged. Any future work from the maintainers around private planned data can use this PR as reference 👍🏻

PR #784, opened 2026-05-11
build: add linux/s390x build target

ansgarm:
## Summary

* Add `linux/s390x` build target for IBM Z platform support
* Add `s390x` to the `goarch` build matrix in `.github/workflows/build.yml` with exclusions for freebsd, windows, and darwin
* Add `linux_s390x.zip` artifact entry to the CRT release manifest


BBBmau (approved):

rainkwan (approved):
LGTM 😄

PR #813, opened 2026-09-10
suppress unused openpgp vulnerability

ansgarm:
## Summary

- Suppress `GO-2026-5932` in the release security scanner.
- The provider does not use `golang.org/x/crypto/openpgp`; `go mod why golang.org/x/crypto/openpgp` confirms there is no package dependency.
- The advisory has no fixed upstream version.

Assisted-by: OpenCode


rainkwan (approved):
Seems reasonable to me

